PR #62, opened 2020-12-06
1st Hostility Tweaks

zumbak04:
## Changelog:
- The Order family considers other families to be Hostile instead of Astray.

# How to test:
Just check a few faiths from the Order family.


TheGroxEmpire (commented):
Changes seems to made it to the game. Most religion outside of Order family are considered hostile except for "evil" religions and those who have order syncretism doctrine. Though I think we should get other Order faiths to consider Secularism (Fortunism and Rationalism) hostile.

zumbak04:
> Though I think we should get other Order faiths to consider Secularism (Fortunism and Rationalism) hostile.

Why? Seculars are really nice guys that don't conflict with others.

TheGroxEmpire (commented):
> Why? Seculars are really nice guys that don't conflict with others.
Secularism is pretty detached from the rest of the Order family. Secularism can still consider the other Order faiths to be astray but I don't think they should be considered astray by other Order faiths for example. Fortunism and Secularism are insignificant to Dragons, Titanforged, etc.

zumbak04:
> Secularism is pretty detached from the rest of the Order family. While Dragons may consider Titanforged faiths or Elves arcane magic to be akin, Secularism is pretty insignificant and nothing alike compared to the others Order Faiths. Secularism can still consider the other Order faiths to be astray but I don't think they should be considered astray by other Order faiths.

Their view on magic is more pragmatic (as well as titan or dragon view on Arcane magic). It's not like a religion, but a philosophy.
Both goblins and gnomes are created by titans.
The Order family is the most tolerant family, a perfect place for secularists with such views.

TheGroxEmpire (commented):
> Their view on magic is more pragmatic (as well as titan or dragon view on Arcane magic). It's not like a religion, but a philosophy.

Yes I agree with this statement, thus why I think Secularism should consider other Order faith Astray. But not the other way around.

> Both goblins and gnomes are created by titans.

In the same sense that Human were derived from Vrykul or Wildhammer Dwarves from Earthen. I think their cultures have deviated far from their origin.

> The Order family is the most tolerant family, a perfect place for secularists with such views.

Yes, the Secularists could stay on Order family. It's just that I find it weird for a Titan Keeper or Dragon Aspect to find Goblin lust for wealth / power (Fortunism) or Gnomish view of practicality Astray.

zumbak04:
> Yes, the Secularists could stay on Order family.

They can't. Or more like they can, but it will cause gameplay problems.

If they're hated by their own family, it means even the dwarven religion of Khaz Modan is more close to the Order family because of the Order Syncretism than gnomes and goblins.

Secularism no matter its tenets won't be able to become closer to the Order Family while everyone else can.

Thus we either separate Secularism from the Order creating a very small family with only one religion and update our religious design. I wouldn't do that.
Or we keep them in the Order Family on very unique rules where they have to accept the syncretism with their own family to be treated like a normal family member. I wouldn't do that.

> Dragon Aspect to find Goblin lust for wealth / power (Fortunism)
That's a result of the tenets. They may reform. In its foundation, they have the same virtues and sins as the gnomes.

TheGroxEmpire (approved):
All things considered then.

PR #118, opened 2021-02-02
Compressed loading screens and flatmap

MemeWizard:
#### Developer changelog:
- Compressed loading screens (~30 > ~3 mb ) and flatmap (~75 > 20 mb) using the right settings
- Deleted empty map .psd from `map_data` folder


zumbak04 (approved):
LGTM!

PR #122, opened 2021-02-03
Orcish Elective Prefer Martial and Prowess

zumbak04:
## Changelog:
- Electors in the **_Orcish Elective_** prefer characters with high martial and prowess.
- **_Corruptism_** has the **_Elemental Syncretism_** tenet so the Horde is less divided.

# How to test:
Play the Horde, pick any candidate, open the Electors tab, check why they prefer different candidates.

### DIFF
--- a/common/religion/religions/fel_group.txt
+++ b/common/religion/religions/fel_group.txt
@@ -7,7 +7,7 @@
 	#Main Group
 	doctrine = doctrine_no_head
 	doctrine = doctrine_gender_equal
-	doctrine = doctrine_pluralism_fundamentalist
+	doctrine = doctrine_pluralism_righteous
 	doctrine = doctrine_theocracy_lay_clergy
 
 	#Marriage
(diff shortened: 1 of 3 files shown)


TheGroxEmpire (commented):
So each point of Martial / Prowess is equal to 1 point of election score and only occur if Martial / Prowess >=10. Imo it needs to have bigger than 1:1 value to score ratio. Other factor are just too big for them than to consider a stronger candidate.
I modified my heir to have 30 Martial / Prowess and most of my vassal still preferred other candidate by like hundreds of points in score discrepancy (I did this after changing all of their religion to Corruptic to eliminate Heretic candidate score malus). 
Unless you are planning on overhauling the scoring down the road. Something like 2:1 ratio I think should be better for now.

zumbak04:
> I modified my heir to have 30 Martial / Prowess and most of my vassal still preferred other candidate by like hundreds of points in score discrepancy (I did this after changing all of their religion to Corruptic to eliminate Heretic candidate score malus).

I made them astray in 404f10321433af041cc2d1c20528ade62586aac8 so the religion should matter less now.

> Unless you are planning on overhauling the scoring down the road. Something like 2:1 ratio I think should be better for now.

This ration is too big IMHO. It means Orgrim will get +100 score just from this.

TheGroxEmpire (approved):
Yeah 1:1.5 was my 2nd ratio choices. So that's fair enough to me. The Astray opinion also works.

PR #131, opened 2021-02-13
Troll Ears

zumbak04:
## Changelog:
- Added a variation of long ears for the trolls, Pulled Back.

## Developer changelog:
- Cleared other ear genes to reset their look.

# How to test:
Feel free to set ear genes you like.


ercarp (approved):
Ears look good and nothing's crashing. I can always adjust the ear genes later for the subspecies of trolls if I feel like some of them should have slightly different-sized ears or something.

Issue #114, opened 2021-02-01
[BUG] | tenet_ritual_cannibalism


MemeWizard:
**Your mod version is:**
4f341c17

**What expansions do you have installed?**
`-`
**Are you using any submods/mods? If so, which?**
`-`
**Please explain your issue in as much detail as possible:**
![image](https://user-images.githubusercontent.com/46576860/106512405-9a5b6700-64e2-11eb-988f-8227ce61f78f.png)

**Steps to reproduce the issue:**

![image](https://user-images.githubusercontent.com/46576860/106512436-a810ec80-64e2-11eb-9a05-a9cb518bfd39.png)

**Upload an attachment below: .zip of your save, or screenshots:**

![image](https://user-images.githubusercontent.com/46576860/106512446-afd09100-64e2-11eb-9999-18c7a3f1fc9a.png)

PR #135, opened 2021-02-13
Fixed bug with Shath'Gral faiths not being allowed Ritual Cannibalism.

PaperPlate-Github:
## Note:
Commented out the line that blocked the religion from being able to use it for custom faiths. I did the same for any other lines with the comment "They have it by default" since I assumed that meant religions under that should be able to use the tenet.

I only commented them out since the issue was vague in what it wanted changed.
#114

## Changelog:
- Custom Shath'gral, Deathbringer, and Burning Legion religions may now use the Ritual Cannibalism tenet.

## Developer changelog:
- The religions were marked as not being able to use the tenet in the core_tenets file.
- There could be other tenets that were marked as unavailable to religions that should be able to use them.


common/religion/doctrines/00_core_tenets.txt
```diff
@@ -1650,7 +1650,7 @@
 		is_shown = {
 
 			# Warcraft
-			evil_public_religion_trigger = no	#They have it by default
+			#evil_public_religion_trigger = no	#They have it by default
```
zumbak04, line 1653: You fixed it wrong. They ALREADY have it from their special doctrine. Thus they shouldn't be able to pick and their religions shouldn't have it.

zumbak04 (changes requested):

common/religion/doctrines/00_core_tenets.txt
PaperPlate-Github: I've re-disabled the tenets.
Cannibalism isn't included in the special doctrine. Should I enable that tenet for evil religions to pick or should I add it to the special doctrine? Or is it ok that they can't pick the tenet when making heresies, and the issue is instead that G'huuni should be given a different tenet?

zumbak04: Nevemind, I think I figured it out. So, there's `doctrine_end_justifies_means` I talked about that all evil religions have, but Ritual Cannibalism isn't included there. Thus we should allow picking Ritual Cannibalism and that's all.

zumbak04 (approved):
LGTM!

PR #140, opened 2021-02-14
Troll Clothes

zumbak04:
## Changelog:
- Every troll except the Drakkari use shoulder-naked robes and animal skins.
- The Drakkari use northern shirts with fur.

# How to test:
Check and tell if the clothes are good.


ercarp (approved):
The clothes look perfect, very fitting for the troll aesthetic. I don't have anything to add. :)

PR #152, opened 2021-02-16
Orcish Beauty

zumbak04:
## Changelog:
- Removed some beauty limitations so they won't break the orcish look.

# How to test:
Tell me if you want to remove more limitations.


ercarp (approved):
Looks good as well! 👍

PR #183, opened 2021-02-21
Undead Trait Fix

ercarp:
## Changelog:
- Removed `base = being_undead_basic` from Nathanos and Sylvanas as their body types should not be affected by the trait.

# How to test:
- Make sure this doesn't break anything.


zumbak04 (approved):
LGTM!

PR #187, opened 2021-02-21
Generated Random Locators

zumbak04:
## Changelog:
- Generated random locators so army, building, siege models are visible and placed on the map now.

# How to test:
Open the game, make sure armies, sieges, buildings have models placed on the map.


TheGroxEmpire (approved):
Unit appeared, moved, attacked, and embarked / disembarked properly. Holdings are also showing appropriate with their government type. Sieges appeared to have catapult in place also.

PR #188, opened 2021-02-21
Ogre Beards and Hairstyles

ercarp:
## Changelog:
- Added a set of beards and hairstyles for the ogre ethnicity
- Added portrait modifier: 80% of ogres are bald and clean-shaven.

# How to test:
- Provide feedback on the chosen beard and hairstyle options added for male and female ogres by generating random ogres in the Ruler Designer.
- Provide feedback on the rate (80%) of bald and clean-shaven ogres. Should it be higher or lower?

### DIFF
--- a/common/ethnicities/wc_ethnicities_ogre.txt
+++ b/common/ethnicities/wc_ethnicities_ogre.txt
@@ -383,10 +383,10 @@ ogre_ethnicity = {
 	
 	###Other	
 	hairstyles = {
-		1 = { name = no_hairstyles 		range = { 0.0 1.0 } }
+		1 = { name = ogre_hairstyles 		range = { 0.0 1.0 } }
 	}
 	beards = {
-		1 = { name = no_beard 		range = { 0.0 1.0 } }
+		1 = { name = ogre_beards 		range = { 0.0 1.0 } }
 	}
 	gene_eyebrows_shape = {
         1 = {  name = no_eyebrows             range = { 0 1 }     }
--- a/common/ethnicities/wc_ethnicities_ogre.txt
+++ b/common/ethnicities/wc_ethnicities_ogre.txt
@@ -386,7 +386,7 @@ ogre_ethnicity = {
 		1 = { name = ogre_hairstyles 		range = { 0.0 1.0 } }
 	}
 	beards = {
-		1 = { name = ogre_beards 		range = { 0.0 1.0 } }
+		1 = { name = orcish_beards 		range = { 0.0 1.0 } }
 	}
 	gene_eyebrows_shape = {
         1 = {  name = no_eyebrows             range = { 0 1 }     }
--- a/common/genes/03_genes_accessories_beards.txt
+++ b/common/genes/03_genes_accessories_beards.txt
@@ -303,5 +303,20 @@ accessory_genes = {
 				1 = empty
 			}
 		}
+		ogre_beards = {
+			index = 15
+			male = {
+				15 = male_beard_indian_03 #
+				5 = male_beard_steppe_02
+				15 = male_beard_northern_03
+				15 = male_beard_northern_02
+				15 = male_beard_northern_01
+				15 = male_beard_western_07
+				5 = male_beard_mena_03
+			}
+			female = {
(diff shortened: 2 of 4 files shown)


TheDrPibb1999:
Every ogre i have generated has had a beard, I think the 80% for clean shaven is a good idea, not too common in the game

ercarp:
> 
> 
> Every ogre i have generated has had a beard, I think the 80% for clean shaven is a good idea, not too common in the game

My bad, I think I wasn't clear on this. The clean-shaven and baldness modifiers aren't taken into consideration when you generate ogres using the Ruler Designer or Debug Portrait Editor. This is something that's calculated in-game. If you load into a game and look for ogres in the game world, you should see the 80% modifier come into action.

TheGroxEmpire (commented):
You're right about needing to make custom hairstyle for Ogres. The Vanilla hairstyle are definitely not workable at all. The hairs are floating and clipping through the horn most of the time:
<details>

image
image
image
</details>

I don't thing this is fixable without changing much of their current shape, but I'm happy to be proven wrong.

zumbak04:
> You're right about needing to make custom hairstyle for Ogres. The Vanilla hairstyle are definitely not workable at all. The hairs are floating and clipping through the horn most of the time:
> 
> I don't think this is fixable without changing much of their current shape, but I'm happy to be proven wrong.

True. So I took a closer look, these hairstyles work well...
`male_hair_northern_01` AKA Very Long Wide
`male_hair_western_04` AKA Fade
`male_hair_indian_01_wc` AKA Bald Top

common/genes/03_genes_accessories_beards.txt
```diff
@@ -303,5 +303,20 @@ accessory_genes = {
...
+				15 = male_beard_northern_03
+				15 = male_beard_northern_02
+				15 = male_beard_northern_01
+				15 = male_beard_western_07
+				5 = male_beard_mena_03
+			}
+			female = {
+				1 = empty
+			}
+		}
```
zumbak04, line 320: These are orcish beards except `male_beard_steppe_02`. Maybe we don't need this gene and the ogres will use the orc beards?
What to do with `male_beard_steppe_02` is up to you. You may include it in the orcish beards.

ercarp: > 
> 
> These are orcish beards except `male_beard_steppe_02`. Maybe we don't need this gene and the ogres will use the orc beards?
> What to do with `male_beard_steppe_02` is up to you. You may include it in the orcish beards.

I added it for a reason. Ogres are canonically supposed to be able to grow hair on their upper lip as well, so I wanted to have one beard like that. As you can see, I made it much rarer to occur than the other beards. Not to mention ogres having a beard, in general, is quite rare.

But I suppose we could just use the orc beards and add `male_beard_steppe_02` into the mix as a rare alternative.

ercarp: Addressed this in f7b06f9492a29674102d6c15e20dc8c9f1a4de7b.

common/ethnicities/wc_ethnicities_ogre.txt
```diff
@@ -383,10 +383,10 @@ ogre_ethnicity = {
```
zumbak04, line 383: What about limiting hairstyles to the ones that don't clip or fly?
https://github.com/Warcraft-GoA-Development-Team/Warcraft-Guardians-of-Azeroth-2/pull/188#pullrequestreview-595418862

zumbak04 (changes requested):

common/ethnicities/wc_ethnicities_ogre.txt
ercarp: I personally don't think clipping is such a big deal (played too much WoW so I'm used to it) but can do. 👌

ercarp: Taken care of in 3cb6fea594db534be0e8c1ebd5ecfd4ffadc4d5f!

zumbak04 (approved):
LGTM!

PR #189, opened 2021-02-21
Important Lore Character

zumbak04:
## Developer changelog:
- Added Plot Armor or `important_lore_character` to indicate who is an important lore character instead of the long list.

# How to test:
Tell me if it looks better and check that I didn't miss anyone.


ercarp (approved):
LGTM, but I think we'll probably have to add the flag to a couple more characters and not just the ones I listed.

The character I listed were ones that were generating random families. Characters with pre-existing children don't usually do this and so none of those were on the list. We should add this flag to characters like Terenas, Varian, Llane, etc. so they get the plot armor effect as well.

zumbak04:
> LGTM, but I think we'll probably have to add the flag to a couple more characters and not just the ones I listed.
> 
> The character I listed were ones that were generating random families. Characters with pre-existing children don't usually do this and so none of those were on the list. We should add this flag to characters like Terenas, Varian, Llane, etc. so they get the plot armor effect as well.

I opened an issue, so anyone can continue once it's merged to the master.
https://github.com/Warcraft-GoA-Development-Team/Warcraft-Guardians-of-Azeroth-2/issues/192

PR #193, opened 2021-02-21
Elf Clothing Sets

ercarp:
## Changelog:
- Added clothing sets to elves based on Vanilla CK3 cultures.
- High Elves, Blood Elves, and Highborne use **_Byzantine_** culture clothes.
- Night Elves use **_Steppe_** culture clothes.
- Elves do not wear crowns.

# How to test:
- Provide feedback on the choice of clothing by starting up the game and observing clothes worn by the aforementioned elves in-game.


Deuce20201 (commented):
For Blood Elves, Highborn and High Elves I like the choice of Byzantine clothes, ya know, all regal and stuff. 

However I can't really say I like Night Elves in Steppe Clothing. Kind of seems off to me.

ercarp:
> 
> However I can't really say I like Night Elves in Steppe Clothing. Kind of seems off to me.

I agree that the colors for some of the Steppe variations are a bit off, but I'm not sure how to fix that. I'm not a fan of the extremely deep blues for example. If there was a way to restrict the colors to be more fitting for the Night Elf aesthetic, I think they would work better.

The Steppe clothes have a slightly Japanese look to them, or at least as close as we can hope to get in Vanilla CK3, which is what the decision was based off of:

> The architecture of the night elves, according to an interview with Chris Metzen, draws its influences from both Nordic and **_Japanese_** building style.

This quote is of course in reference to architecture but we do see some of that same inspiration on clothing in my opinion. That said, I'll wait for a few more reviews and look into if it's possible to restrict some of the color choices on their clothing somehow.

Deuce20201:
> > The architecture of the night elves, according to an interview with Chris Metzen, draws its influences from both Nordic and **_Japanese_** building style.
> 
> This quote is of course in reference to architecture but we do see some of that same inspiration in their clothing in my opinion. That said, I'll wait for a few more reviews and look into if it's possible to restrict some of the color choices on their clothing somehow.

Ah, I did not know that they based NE off of Nords and the Japanese. I guess in that sense the Steppe clothing is a good choice.

TheDrPibb1999 (approved):
For base ck3 clothing I think the current clothing for the respective elves works quite nice, I also believe the circlet crowns work quite well for the high elves, but in time I think some custom clothing/headwear will make it all the better. but my only suggestion is I believe lor'themar has never worn a crown ever and he has his nice high pony tail, so maybe in that 905 start make that adjustment? other then that nice placement :)

ercarp:
> 
> 
> For base ck3 clothing I think the current clothing for the respective elves works quite nice, I also believe the circlet crowns work quite well for the high elves, but in time I think some custom clothing/headwear will make it all the better. but my only suggestion is I believe lor'themar has never worn a crown ever and he has his nice high pony tail, so maybe in that 905 start make that adjustment? other then that nice placement :)

Lor'themar does not have custom DNA yet, but I'll try my best to give him the appropriate haircut. Will also give him a `no_hat/no_headgear` flag if possible. I'll make his portrait soon™ if Garthor doesn't get to it before me.

ercarp:
Changed the color of Low and High Nobility Steppe clothes to be more Night Elf appropriate.

image
image
image
image
image
image

</details>

Changing the colors of commoner clothing is a little more convoluted, so I haven't done that yet. It's also difficult to figure out which parts of the color palette correspond to which parts of the clothing, so for the time being I've made the entire color palette mostly blues, purples, and silvers.

zumbak04 (approved):
LGTM!

PR #196, opened 2021-02-22
Gnome Ethnicity Localization

ercarp:
## Changelog:
- Added localization for the Gnomish Ethnicity.
- Also cleaned up the file a little bit so it's easier to tell where everything is.


zumbak04 (approved):
LGTM!

PR #202, opened 2021-02-23
3rd Bug Crusade

zumbak04:
## Changelog:
- Fixed an issue where Lordaeron could build a Grand Mosque instead of Grand Cathedral.

## Developer changelog:
- Replaced `intellect_good_1` by `shrewd`.

# How to test:
Check the bug mentioned in the changelog is no more.
If you can, check the Files changed on Github.


TheGroxEmpire (approved):
Can confirm that country of Lordaeron could building Cathedral instead of Mosque now.

PR #213, opened 2021-02-27
Scripted Appearance

zumbak04:
## Changelog:
- Added a beauty trait that doesn't affect the character's look.
- Muscles of lore characters don't change depending on their prowess.

# Note:
To make the character not affected by `beauty_good` traits, you have to replace their `beauty_good` trait with `beauty_good_lore` like it's done with Sylvanas in the example.

# How to test:
Tell if you're ok with changes and they work fine.


ercarp (changes requested):
For some reason, the game seems to still alter Sylvanas's jaw and chin.

On the left is the DNA I made shown in Debug Portrait Editor. On the right is the version we see in-game. Pay attention to her chin and jawline.

image
image

</details>

The difference is quite subtle, so if this can't be fixed easily, I'd probably just let it be for now.

zumbak04:
> For some reason, the game seems to still alter Sylvanas's jaw and chin.

> The difference is quite subtle, so if this can't be fixed easily, I'd probably just let it be for now.

The question, how you added the trait?

ercarp:
> The question, how you added the trait?

I didn't add the trait to her. I assumed you had already added the trait to Sylvanas's historical entry as I was able to use `remove_trait <ID> beauty_good_lore` on her in-game.

Also, I just checked the history file to confirm this is the case. 😕

zumbak04:
> I didn't add the trait to her. I assumed you had already added the trait to Sylvanas's historical entry as I was able to use `remove_trait <ID> beauty_good_lore` on her in-game.

How you got the two versions then?

zumbak04:
> I didn't add the trait to her. I assumed you had already added the trait to Sylvanas's historical entry as I was able to use `remove_trait <ID> beauty_good_lore` on her in-game.
> 
> Also, I just checked the history file to confirm this is the case. 😕

Fixed in 034aafb48a7cb0d7c1a3788fb47c3af49871d0e2.

ercarp (approved):
Sylvanas is no longer altered by the trait in-game. LGTM!

PR #235, opened 2021-03-15
Skin Value

zumbak04:
## Changelog:
- Added the skin value slider.

# How to test:
We need to keep only one slider pick one. I'm for the Value. Here's why:

There are Hue, Saturation, and Value.

<details>
<summary>Saturation Pallete</summary>

image

</details>
<details>
<summary>Value Pallete</summary>

image

</details>

Basically, our current palette is combined out of these 2 pallets.

And thus there's a problem: when the value is set to 0.5 and more, it doesn't affect the upper part of our pallet because it's maxed out there. The saturation is vice versa, 0.5 and more doesn't affect the bottom part.

So why would I prefer the value? Most of our skin colors aren't dark and picked from the upper part and the Value affects it the most creating the dark version of not-saturated colors like the greenish color of the gray orcs.

<details>

image

</details>

Test it and tell your opinion.

### DIFF
--- a/common/ethnicities/00_ethnicities_templates.txt
+++ b/common/ethnicities/00_ethnicities_templates.txt
@@ -89,6 +89,12 @@ ethnicity_template = {
 		# Black
 		30 = { @brown_color_x_min @black_color_y_min @brown_color_x_max @black_color_y_max }
 	}
+	gene_skin_saturation = {
+		1 = { name = skin_saturation    range = { 0.5 0.5 }    }
+	}
+	gene_skin_value = {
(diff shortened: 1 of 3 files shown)


ercarp (approved):
+1 for the Value slider as well. The Saturation slider doesn't work the way I thought it would as you have to go very dark on the hue palette to get the kind of results I was going for, so it seems like Value is the better option.

This is a very good addition as it will give us some more variety in skin colors.

TheDrPibb1999 (approved):
+1 Value, I agree, it does get better results for skin tones that the saturation pallete cannot acquire.

PR #340, opened 2021-04-20
Mountain Dweller modifier

Gaben1025:
- GDC if useless

## Changelog:
- Adds modifier to Gnomes/Dwarfs/Troggs living in Khaz Modan, so that their dev isn't because they live in a mountain province.

## Dev Changelog:
- Added a trigger in the game_start and the on_title_gain files to mention the new effect

## Need to add:
- Pick and choose the provinces mentioned more accurately.
- Add more cultures to get the modifier, like Black Iron and so on.
- Get the modifier to work...

### DIFF
--- a/common/modifiers/00_province_modifiers.txt
+++ b/common/modifiers/00_province_modifiers.txt
@@ -110,5 +110,5 @@ winter_harsh_modifier = {
 #Warcraft
 mountain_dwellers_modifier = {
 	icon = county_modifier_development_positive
-	development_growth_factor = .25
+	development_growth_factor = 0.25
 }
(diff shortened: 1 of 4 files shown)


common/modifiers/00_province_modifiers.txt
```diff
@@ -106,3 +106,9 @@ winter_harsh_modifier = {
 	defender_winter_advantage = 5
 	hard_casualty_winter = -0.2
 }
+
+#Warcraft
+mountain_dwellers_modifier = {
+	icon = county_modifier_development_positive
+	development_growth_factor = .25
```
MemeWizard, line 113:
```suggestion
	development_growth_factor = 0.25
```

Gaben1025:
`this = { completely_controls = title:d_central_dun_morogh }`  Needs to be changed to mention the whole culture instead of just the player

PR #345, opened 2021-04-22
Less Magic Values

zumbak04:
## Developer changelog:
- A possible way to shorten magic values.
- Approved magic works as multiplayer now.

# How to test:
Tell me your opinion. If it works, I extend it to other values.


MemeWizard:
> Tell me your opinion. If it works, I extend it to other values.

Should work.

PR #372, opened 2021-05-09
Masonry Faith

zumbak04:
## Changelog:
- Added a faith for the Defias Brotherhood, Masonry. It's freer than the Holy Light and deposes theocracies.

# How to test:
The Defias Brotherhood exists in the 603 and 605 bookmarks, check their faith there and tell your opinion.


Grimlawck (approved):
Looks good to me I see no problems with it. I feel like the things they believe in fit them well.

PR #413, opened 2021-06-04
Main Menu

zumbak04:
## Changelog:
- Added the art to the main menu and removed the standing models from there because our static models will look strange there.

# How to test:
Load into the main menu. Tell your opinion about the composition and stuff.


DatFrenchBanana (approved):
Works perfectly as intended on my MAC.

PR #420, opened 2021-06-06
White-haired Dal'rend

zumbak04:
## Changelog:
- Dal'rend is forced to have white hairs in the 603 bookmark.

# How to test:
Take a look and tell your opinion.


icelille999 (approved):
Good idea!

PR #426, opened 2021-06-10
Missing Bracket Hotfix

zumbak04:
## Changelog:
- Added a missing bracket that made beards of other scripted characters to disappear.

# How to test:
Confirm other scripted characters have beards. Should use 1.3.1 version.


icelille999 (approved):
Tested. Thanks for catching and fixing this.

PR #434, opened 2021-06-12
Relative Old

zumbak04:
## Developer changelog:
- Even the elves can have the Old nickname now because we use relative age: `age_45_value`, etc.
- Moved `WCNIC.1` to `three_year_playable_pulse` for performance reasons: it fires only for playable rulers.
- Made `WCNIC.1` rarer.

# How to test:
Check the changes, tell if you approve them.

### DIFF
--- a/common/on_action/birthday.txt
+++ b/common/on_action/birthday.txt
@@ -6,11 +6,6 @@ on_birthday = {
 		on_birthday_childhood
 		on_graceful_aging_birthday
 	}
-	# Warcraft 
-	random_events = {
-		65 = 0
-		35 = WCNIC.1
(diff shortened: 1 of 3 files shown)


ElMariuso:
> Moved WCNIC.1 to three_year_playable_pulse for performance reasons: it fires only for playable rulers.
So the AI ​​will not be able to have this nickname suddenly?

The rest looks good to me.

zumbak04:
> So the AI ​​will not be able to have this nickname suddenly?

No, the AI is able to get the nickname.

ElMariuso (approved):
So, approved.

PR #447, opened 2021-06-15
Bald Bun and Hood Clipping

zumbak04:
## Changelog:
- Fixed an issue where Bald Bun clipped with Hood.

# How to test:
Open the Barbershop, make a character with Bald Bun and Hood as headwear.


CatWitchLily (approved):
Works as expected. I tested many species and ages and did not see any clipping.

PR #490, opened 2021-06-28
innovation_prestige_value

zumbak04:
## Developer changelog:
- Used a value for innovation prestige discount instead of long and confusing triggers.

# How to test:
Confirm it works.


Gaben1025 (approved):
LGTM!

PR #498, opened 2021-06-30
Single Charge Position

zumbak04:
## Developer changelog:
- Changed all single charge CoA positions so we no longer need to raise our CoAs a bit higher to be centered. This is an updated mask:
<details>

image

</details>

# How to test:
Check our CoAs positions, tell if they're centered still.


MemeWizard:
The most are fine, but some are a bit higher than it feels they should be.

<details>

image
image
image
image
image
image
image

</details>

runtd (approved):
Everything seems alright for me

zumbak04:
> The most are fine, but some are a bit higher than where it feels they should be.

Check in 85f6a128ea1363b93676967bbf3dc2956cea5f8c.

MemeWizard (approved):
b

PR #501, opened 2021-07-02
Limited Draenor Levies

zumbak04:
## Changelog:
- Limited Draenor raider levies. They can't go over 4000 (was 8000 before).

# How to test:
`event WCDRA.200` to force the raider event on you.


CatWitchLily (approved):
I spammed the event many times and did not get any army over 4k.

zumbak04:
> I spammed the event many times and did not get any army over 4k.

Wow, thanks for the review!

Samio321 (approved):
I spammed the event as well as Stormwind and got under 4000 levies on the enemy everytime. Interestingly though if I gained land before firing the event the levies would often be above 4000. This is almost certainly fine anyway in terms of balance so I'm giving it an approve.

PR #503, opened 2021-07-03
Satyr, Dryad Immortality

zumbak04:
## Changelog:
- Satyrs, dryads aren't immortal now because it forced their child don't age so they stayed child forever.

# How to test:
Observe satyr and dryad children.


riandaniar (approved):
Both Satyr and Dryad now age accordingly.

PR #513, opened 2021-07-07
Menopause and Andropause

zumbak04:
## Changelog:
- Added On Menopause for females and On Andropause for males it appears if:
  - The character is a long-lived race and has 5 children for females and 7 children for males.
  - The character has reached the age of 45 for females and 60 for males (long-lived races have scaled equivalents.

# How to test:
Observe the game. Check if the trait is added right and the AI doesn't marry infertile characters.


TheDrPibb1999 (commented):
Seems kinda odd that long-lived races can only have a few kids, will that make the "Fecund" trait completely irrelevant?

zumbak04:
> Seems kinda odd that long-lived races can only have a few kids

Otherwise, dwarves can have 100 children by the end of their lives (1 child per 3 years).

> will that make the "Fecund" trait completely irrelevant?

No, Fecund is genetic thus you have it from birth.

TheDrPibb1999 (commented):
Just as a personal opinion i think the age requirements are well set, but I think 3-4 children is a bit low, 5-7 would be my input or even spacing out the duration between the avability to get pregnant, as least for women, men can be fertile for a long long time and have many children with many females (bastards) , but other than that the trait looks well

zumbak04:
> 5-7 would be my input

Changed in de3aaa70080b53304bad46667581aba7248dc9cb.

TheDrPibb1999 (approved):
LGTM

CantTouchThisJinTao (approved):
Everything seems to be working as intended, I think 5-7 is a decent number for the child limit (that or 5-6).
Looks good.

PR #526, opened 2021-07-11
try_to_set_race_effect Memory Leak

zumbak04:
## Developer changelog:
- `try_to_set_race_effect` took too much space in RAM, 2 GB. Putting it in an event fixed a problem.

# How to test:
Run the CK3, open the [Task Manager](https://en.wikipedia.org/wiki/Task_Manager_(Windows)) to check the RAM. Upon the main menu loading, CK3 should use about 3GB of RAM.

<details>

image

</details>

### DIFF
--- a/common/casus_belli_types/00_religious_war.txt
+++ b/common/casus_belli_types/00_religious_war.txt
@@ -3139,7 +3139,7 @@ undirected_great_holy_war = {
 
 				# Warcraft
 				scope:title_recipient = {
-					try_to_set_race_effect = { FLAG = no_gene }			#Assigns race trait
+					trigger_race_giving_no_gene_effect = yes			#Assigns race trait
 				}
 
 				scope:the_faith.great_holy_war = {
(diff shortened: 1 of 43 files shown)


Remerod:
I think mine uses more? Here a pic of menu
image
One when the game loads 
image
And one when the game is closed
image

I'm not really the guy to ask about performance and such, but since you said it was urgent i tried it out ;)

CatWitchLily (commented):
On the menu, my game uses 3.8gb of ram, which is a bit more than 3gb. If this difference is expected then I have no other problems with it.

Remerod:
Ok, new try, here some new screenshots of just CK3 Warcraft
On startup
image

In Menu
image

Remerod (approved):
For me it uses about 3gb ram as I have shown with my pics that i shared. So it works for me

robmart (approved):
Master (Main menu): 
image

Master (in-game);
image


Fix (main menu):
image

Fix (in-game):
image

Not quite reduced to 3GB for me, but significantly reduced regardless.

zumbak04:
> On the menu, my game uses 3.8gb of ram, which is a bit more than 3gb. If this difference is expected then I have no other problems with it.

It's fine.

robmart:
@CatWitchLily Can you approve it so it can be merged?

PR #535, opened 2021-07-17
Holy Order Debug

zumbak04:
## Developer changelog:
- Added a decision to debug holy orders, `CREATE HOLY ORDERS`.

# How to test:
Confirm it works, check the code for a reference. Debug decisions are nice.

### DIFF
--- a/common/decisions/wc_debug_decision.txt
+++ b/common/decisions/wc_debug_decision.txt
@@ -1,0 +1,56 @@
+﻿create_holy_orders_debug = {
+	picture = "gfx/interface/illustrations/decisions/decision_major_religion.dds"
+
+	is_shown = {
+		debug_only = yes
+	}
+
+	is_valid = {
+	}
+	effect = {
+		every_independent_ruler = {
+			add_to_list = holy_order_creators
+			
+			every_vassal = {
+				add_to_list = holy_order_creators
+			}
+		}
+		
+		# Creates 3 holy orders
+		every_in_list = {
+			list = holy_order_creators
+			
+			add_character_flag = { flag = holy_order_debug days = 1 }
+			
+			if = {
+				limit = {
+					can_execute_decision = create_holy_order_decision
+					
+					faith = {
+						trigger_if = {
+							limit = { exists = var:created_holy_orders }
+							var:created_holy_orders < 3
+						}
+						trigger_else = {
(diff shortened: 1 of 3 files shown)


common/decisions/wc_debug_decision.txt
```diff
@@ -0,0 +1,56 @@
...
+					}
+				}
+				
+				execute_decision = create_holy_order_decision
+			}
+		}
+	}
+
+	ai_check_interval = 0
+}
```
SolTsuki, line 56: The button works and creates the number of holy orders, the problem comes with the name picking:
image

After testing a lot I can see there are some odd behaviours and the problem is... I'm not sure if it's your code or paradox', because you call create_holy_order_decision and it doesn't have that many changes so it should be ok. This is what I did:
- I tested with the holy light holy orders, I would click on the CREATE HOLY ORDERS decision and it would pic the first holy order and maybe the last holy order, then it would use some generic holy orders that wasn't specified there (somenthing like "Holy Light company of the Holy Light")
- After a while of trying and playing got tired and increased holy orders generated to 6 and did this:
image
 And the result would be something like:
image

- With this:
image
I got:
image
- With holy orders ordered in the same way you get always the same holy orders in the game, also, after comparing I found out that it ignores the holy orders in a certain position and gets always the same holy orders. I think at some point it tries to create holy orders, can't, but the name is already used, somehow. The arrows indicate which holy orders get picked:
image

SolTsuki (changes requested):
Summary: it creates the 3 holy orders for each faith (if possible, if there are not enough rulers with the possibility to do it it won't happen, or maybe you get 2 or 1), but there are some issues with the names picked.

common/decisions/wc_debug_decision.txt
zumbak04: > but the name is already used, somehow.

Yes, I suppose you play the Holy Light faith and there are 2 branches. There's the dwarven branch of the Holy Light that takes these names.

SolTsuki: Yep, that was it, checked both Holy Light branches and all holy orders names are being used, LGTM!

SolTsuki (approved):
LGTM! 3 Holy Orders are created for each faith if possible and all holy order names are used

PR #544, opened 2021-07-21
Blood Troll Flavorization

ercarp:
## Changelog:
- Gave female blood trolls title flavorization for every realm tier:
  - **County**: Warmother _(suggested by @Remerod)_
  - **Duchy**: Grand Ma'da
  - **Kingdom**: Queen Ma'da _(suggested by @Remerod)_
  - **Empire**: Divine Ma'da _(suggested by @Remerod)_
- Flavorization is only active if the character has the **_G'huuni_** faith and **_Nazmani_** culture.
- Disabled titles from male blood trolls completely (this applies to every tier).

# How to test:
- Give feedback on the chosen title flavorization. Should any of them be changed?

### DIFF
--- a/common/flavorization/wc_title_holders.txt
+++ b/common/flavorization/wc_title_holders.txt
@@ -4,7 +4,7 @@
 @county_cultural_prio = 18
 @barony_cultural_prio = 7
 
-blood_troll_patriarch_male = {
+blood_troll_male = {
 	type = character
 	gender = male
 	special = holder
@@ -14,10 +14,44 @@ blood_troll_patriarch_male = {
 	religions = { ghuun_worship }
 }
 
-blood_troll_matriarch_female = {
+blood_troll_count_female = {
 	type = character
 	gender = female
 	special = holder
+	tier = county
+	priority = 400
+	top_liege = no
+	cultures = { blood_troll }
+	religions = { ghuun_worship }
+}
+
+blood_troll_duke_female = {
+	type = character
+	gender = female
+	special = holder
+	tier = duchy
+	priority = 400
+	top_liege = no
+	cultures = { blood_troll }
+	religions = { ghuun_worship }
+}
+
+blood_troll_queen_female = {
+	type = character
+	gender = female
+	special = holder
+	tier = kingdom
+	priority = 400
+	top_liege = no
+	cultures = { blood_troll }
+	religions = { ghuun_worship }
+}
+
+blood_troll_empress_female = {
+	type = character
+	gender = female
+	special = holder
+	tier = empire
 	priority = 400
 	top_liege = no
 	cultures = { blood_troll }
--- a/common/flavorization/wc_title_holders.txt
+++ b/common/flavorization/wc_title_holders.txt
@@ -4,6 +4,26 @@
 @county_cultural_prio = 18
 @barony_cultural_prio = 7
 
+blood_troll_patriarch_male = {
+	type = character
+	gender = male
+	special = holder
+	priority = 400
+	top_liege = no
+	cultures = { blood_troll }
+	religions = { ghuun_worship }
+}
+
+blood_troll_matriarch_female = {
+	type = character
+	gender = female
+	special = holder
+	priority = 400
+	top_liege = no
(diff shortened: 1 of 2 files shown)


Remerod (commented):
So on the Topic on Male Blood Trolls. I dunno if this is the case any longer but in Early WoW Troll lore, troll women was not valued the same as troll males, they could be sold to other tribes and did not choice their mate's and such. Most notable trolls where males. with trolls being a popular playable race Blizzard in my opinion tried to do a better job representing female trolls and went silently away from the "no rights" female trolls. But the core remain the same Trolls are an old/early civilization themed with men in control. 
So the thing with Blood Troll that was so strange compared to other trolls was the gender role are reverse. this is both ironic and could seen as some sort of corrupted theme of "standard" troll society (or just divine irnony/justice) One way Blood trolls are described are such as:
Their society is also matriarchal; males follow the orders from females without question lest they get chained up and punished or outright sacrificed. There is an example that question this theme in the lore, and that is Zul, with is later reborn and implied to be/become blood troll. but even if this is the Case Zul culturally was a Zandalari and would rather be an exception rather then a rule.
So in my opinion, male Blood Troll should not be able to be granted "any" landed titles whatsoever.

on the topic on titles holders, this is how it looks to me, it seemed to be in order?
<details>

image

</details>

I checked around in game, and both Count and Duke holders seem to be named Grand Ma'da
I have some suggestions for some changes to this;
Count: Great/High/Tribe Ma'da or Warmother (Ma'da is used for most females so perhaps something singnifi the rules of a tribe/village. The Leader of the Amaki tribe is called Warmother Nagla, also the leader of the bloodhunter's is called Warmother Boatema.)
Duke: Grand Ma'da (as it is now)
King: Queen Ma'da (there is a blood troll called "Poz'ga the Butcher Queen" so Queen is in their vocabulary)  
Emperor: Divine/Grandest/(Grand Queen)/(Blood Queen) Ma'da (Rastakhan is titled God King, so a title that implies to be chocen among the loa's, but perhaps something more.. G'huuni would be better here)

localization/english/culture/wc_culture_titles_l_english.yml
```diff
@@ -26,4 +26,7 @@
  principality_nightborne:0 "Principality"
 
  lord_nightborne_male:0 "Lord"
- lady_nightborne_female:0 "Lady" 
\ No newline at end of file
+ lady_nightborne_female:0 "Lady"
+
+ blood_troll_patriarch_male:0 "#"
```
zumbak04, line 31:
```suggestion
 blood_troll_patriarch_male:0 ""
```

common/flavorization/wc_title_holders.txt
```diff
@@ -4,6 +4,26 @@
 @county_cultural_prio = 18
 @barony_cultural_prio = 7
 
+blood_troll_patriarch_male = {
```
zumbak04, line 7: You should define a tier. Otherwise, every vassal will be called Ma'da.
<details>

image

</details>

zumbak04 (changes requested):

localization/english/culture/wc_culture_titles_l_english.yml
ercarp: Are you sure about this? Leaving it empty adds an unsightly empty space before their name. With the `#` added, there's no space and it looks more pleasing to the eyes.

image

common/flavorization/wc_title_holders.txt
ercarp: Added in 1b9ca25949cda89b2be8502e5f286054cdf88f70.

Remerod (approved):
So i didn't check/look on the Queen Ma'da and Divine Ma'da ingame, but i think this is good, i'm a little bit uncerten on Divine (even though i suggested it) becouse it dosent really sound Ghuuni but I can't think on anything better so i think it works for of now. 
I think Warmother was a really good choice for count, it looks nice. 

<details>

image

</details>

so overall i'd say this is nice and believable flavor for the culture/religion.

zumbak04 (approved):
LGTM!

PR #548, opened 2021-07-22
Horns Clipping with Hairs

zumbak04:
## Changelog:
- Fixed draenei horns clipping with hairs.


CantTouchThisJinTao (approved):
Looks perfect :)

PR #555, opened 2021-07-24
decal_apply_order is Back

zumbak04:
## Developer changelog:
- Returned `decal_apply_order` since markings are indeed pained on the skin, not under.

# How to test:
Tell if markings look good still.


ercarp (approved):
They still look good, or even better I would say. LGTM!

PR #558, opened 2021-07-27
High Elf Markings

ercarp:
## Changelog:
- Added **6** facial markings to be used by **Knights** and **Commanders** with the **_Quel'dorei_** culture.
  - This includes **4** unique markings, including left and right variations of markings 02 and 03.

## To-do:
- [ ] Marking `facial_highelf_markings_04` is only to be used by clean-shaven male characters.
- [ ] Fix the issue with `trait = creature_high_elf` not working.
- [ ] Create `facial_highelf_markings_05`.
- [ ] Create `facial_highelf_markings_06`.

## Known issues:
- The `trait = creature_high_elf` part of the modifier does not seem to work. Perhaps it needs to be changed to `has_trait` instead.


zumbak04:
> * The `trait = creature_high_elf` part of the modifier does not seem to work. Perhaps it needs to be changed to `has_trait` instead.

Yes, this is an issue though we define clothes by culture (since a high elf doesn't have to have the high elven culture).

common/genes/wc_gene_facial_markings.txt
```diff
@@ -440,6 +440,149 @@
 				boy = male
```
zumbak04, line 440: Localized markings in the Ruler Designer and Barbershop.
<details>

image

</details>

zumbak04 (changes requested):

PR #559, opened 2021-07-27
Naga Belly Color

zumbak04:
## Changelog:
- Naga use hair color on their belly.

# How to test:
Tell if it looks good.


common/genes/wc_gene_body_naga_test.txt
```diff
@@ -45,6 +45,36 @@ morph_genes = {
 					}
 					priority = 1
 				}
+				decal = {
```
kaldrskegg, line 48: Can we do this for scales as well on the back and back of head?

kaldrskegg (changes requested):
Wondering if we can do this for the back scales also? But I would wait a bit to approve because the masks might have to change as I am working on the diffuse right now.

common/genes/wc_gene_body_naga_test.txt
zumbak04: > Can we do this for scales as well on the back and back of head?

Which areas?

kaldrskegg: May be unecessary. In WoW you end up with 4 colors: belly, skin, scales and webbing.
image
Like here, webbing pink, scales green, belly red, skin orange. If its not doable then its fine.

kaldrskegg (approved):
LGTM

PR #580, opened 2021-08-07
Succubi Ethnicity

zumbak04:
## Changelog:
- Added the Succubi ethnicity.
- 7c4a36f2f919ba5ad6be8a99411576bb20e58407: Added Succubi wings for female Sayaadi.
- 1a08b23f83174405b6d95f3b115beab61f0e7826: Added four facial markings.
- 8c436406d7c18ce7996b0ddee73c286ece950094: Added arm marking.

## Tasks:
- [x] @ercarp should visit this branch and edit their faces and hairstyles.
  - [x] Readjust `succubus_facial_marking_01`. _(Dec 19 2021)_
  - [x] Readjust `succubus_facial_marking_02`. _(Dec 19 2021)_
  - [ ] Readjust `succubus_arm_marking`.
- [x] @Eschyles should visit this branch and add normals and properties for the succubi leg texture.

# How to test:
Find the Sayaadi ethnicity in the Ruler Designer.
Find characters with the Sayaadi racial trait via the Character Finder.
Give your feedback.


Remerod (changes requested):
So i had a good look at the Saayadi 
<details>

So first of all, i really like it, it is kinda good, the things i will point out is mostly nit-pick stuff, but i'll guess i start with the biggest first.

**Old look**
image
In the global search for Sayaad at least a third or a fourth of them where old looking, i assume this is not supposed to be this way and that they are all supposed to look young? they are succubus and incubus after all, but there be a lot more people with a grand pa & ma fetish then i would now :P
These wrinkles are the primary reason I **requested a change** 

 **Those legs**
image
I do like the legs! if I could point out anything that would improve it, it would by adding some horns and whiskers.
image

**tail**
image
As you see on these two pictures (and most pics of succubusses) they have stuff on their legs. I don't think they'll be unplayable without them but it would add to their looks.

image
The position of the tail looks weird.
And to be fair it does so on several pictures of Sayaad.
image

So I guess it is in the right position? I guess i would appreciate it if we could give it a little uplift. So the tail ain't so slant. 
image
image

**Arm Mark**
A thing that would add to the Sayaad is their arm marking, so if we could add that it would be great!
image
As you can see it is really similar in colour and texture as the end of the leg one.

**Hair**
I did like the hair cuts of the Male, less so on the females, but that's more of a we don't have the typical hair of Sayaad in basic ck3. So i think it's fine.

**Eye Marking Suggestion**
In wc3 reforged the succubus units have special eye markings, this make them look special and seducing? If we want to add something more that would differentiate them from others we could add this.
_Succubus_
image
_Vile Tormentor_
image
_Vile Temptress_
image
_Maiden of Pain_
image
_Queen of Suffering_
image



**Summary**
Overall i think this Sayaad pull is good, their colours, their eyes. all good. The thing i really want to change primarily is the old look the minority but a lot of them had. 
The others would be good to, but i can understand them not being worked on either if that's a to much hassle.

</details>

ercarp:
> A thing that would add to the Sayaad is their arm marking, so if we could add that it would be great!

I will add the arm marking.

> In wc3 reforged the succubus units have special eye markings, this make them look special and seducing? If we want to add something more that would differentiate them from others we could add this.

I like these a lot. I will add them as well.

ercarp:
Added facial markings and the arm marking. I did not add the Vile Temptress marking because it is too similar to the Succubus marking you showed. But otherwise everything else was added.

CantTouchThisJinTao (approved):
Sayaadi look amazing, not noticing any issues. Not much to say tbh, I think the variety is good, wing colour differences can be hard to notice at times (most likely because of the darkness of the page itself when a character is selected - probably not much that can be done about that - it fits the theme for demons anyway so I wouldn't want it changed). Can definitely notice the difference when looking via the character finder.
All in all no complaints. 10/10 would marry sayaadi.

ercarp:
> One marking though looks less good as the others, the Sayaadi Temptation. image

@Remerod Fixed in 6fbfe2fc335bcdd028eaaa4837aa830513023dcf.

image

Due to how the ethnicity's faces are shaped (high cheekbones), the markings appear smaller than they actually are intended to be. But I don't think it's worth making them longer than intended because then they will look strange on other ethnicities.
image

</details>

Remerod (approved):
@ercarp I very much like the new change for the temptation eye mark, it looks a lot better! :)

Overall I think the Sayaad ethnicity now look very distinct from other species, and very good! Well done everyone!

And unless you really wish to change the arm markings @ercarp I think they are ready to be implemented, well I approve at least

<details>

image

</details>

Melnerag (commented):
### Arm Markings
They work! I also didn't realize they were a thing until I saw them here, and double-checked with the game.

<details>
<summary> Marking </summary>

image
</details>

### Redder Skin
Compared to my earlier screenshots (Oct 29th), redder skin option seems to appear (see: 3rd incubus in screens)

<details>
<summary> Succubus Diversity </summary>

image
</details>

PR #589, opened 2021-08-13
Infusion Modifiers

zumbak04:
## Changelog:
- Assigned modifiers to Infusion traits: Undead, Demon, Void Being, Nature Being, Order Being, Light Being.

# How to test:
Open the game, open the Encyclopedia (`F9` hotkey), open `Traits` tab, search for the mentioned traits and their effects.

Tell if you want to rebalance it or give new effects.

### DIFF
--- a/common/traits/wc_infusion_traits.txt
+++ b/common/traits/wc_infusion_traits.txt
@@ -11,7 +11,7 @@
 			desc = trait_being_undead_desc
 		}
 	}
-	shown_in_ruler_designer = yes
+	shown_in_ruler_designer = no
 	
 	triggered_opinion = {
 		parameter = being_undead_illegal
@@ -23,6 +23,10 @@
 		opinion_modifier = being_undead_intolerant
 		ignore_opinion_value_if_same_trait = yes
 	}
+	
+	diplomacy = -1
+	prowess = 2
+	dread_baseline_add = 15
 	
 	life_expectancy = 25000	# Instead of immortality because immortality locks you from aging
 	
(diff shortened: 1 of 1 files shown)


Remerod (commented):
I think it's weird that an Undead character would be better then diplomacy then a demon, I mean demons lure mortals all the time, while undead.. just are scary. I think at least the Undead and Demon trait should have the same Diplomacy penalty, or switch them around.
also i think demons should have more dread or undeads less, undead can still have more dread but they shouldn't have almost the double dread factor then what a demon have. and i kinda get that all these are kind of a balance factor, but it's weird that undead's are more scary while at the same time also better then diplomacy then demons.
<details>

image

</details>

the other traits get +25000 life expectancy (aka immortal), while **Void Being**  & **Light Being** only get 130 years, is this an oversight or intended?
<details>

image

image

</details>

zumbak04:
> the other traits get +25000 life expectancy (aka immortal), while **Void Being** & **Light Being** only get 130 years, is this an oversight or intended?

Intended since mantids don't live that long.

> also i think demons should have more dread

Then what minus should the demons have?

> undeads less

Then what should the undead have?

Remerod:
What if undead have -2 diplomacy and demons -1, but Undead get +2 in either intrigue or martial, because they are kinda sneaky and they basically fight/guard all the time.
Demons could have +10 in dread while Undead have +12, undead's would still have more but the difference would not be twice as much. 
How does that sound? balancing is not really my strong suit, but this might fit thematically more? i dunno

Remerod (approved):
had a look on the demon and undead traits, it looks a lot better and more balanced, lore wise at least in my opinion.
the only thing now is that light have prowess of +4 while no one has +3, most have +2. This is not a thing that have to change, it is okay in my opinion since i feel most "warriors of light" usually are pretty strong but fewer then the bad guys but i dunno, it is at least something to think about. 

But over all l like this and think it is ok.

PR #607, opened 2021-08-22
Mustache with Goatee

zumbak04:
## Changelog:
- Added a new beard, Mustache with Goatee. Human characters may have it.

# How to test:
Check the beard in the Ruler Designer, Barbershop and in-game and tell your opinion.


Remerod (commented):
So I took a look at this
<details>

I think that this look's okay and are workable for someone like Shaw.
image

image
But Clearly Shaw's mustache doesn't go much at all under his over-lip. witch our version does.

image
If we want it to look more like Shaw Mustache i would suggest to look more like something along the red line along the Mustache like this, and also to make his goatee a little wider.

But overall i think your version works

</details>

zumbak04:
> If we want it to look more like Shaw Mustache i would suggest to look more like something along the red line along the Mustache like this, and also to make his goatee a little wider.

Doing it for @ercarp so I wait for his response.

ercarp (approved):
I tested it using Shaw's DNA and to me it looks perfect. Thank you! ❤️

image
image

</details>

PR #609, opened 2021-08-22
Glowing Iris

zumbak04:
## Changelog:
- Added a new eye type, Glowing Iris.

# How to test:
Check it in the Ruler Designer, tell your opinion.


ercarp (approved):
I like them a lot. Initially, I thought the irises should probably have a dark outline like Glowing Pupils do, but I tested out what that would look like in Photoshop and it didn't look as good as these. Not having the outline makes them look brighter and more glowy somehow. So great job!

PR #652, opened 2021-09-03
Normal and Property Textures Resavin

zumbak04:
## Developer changelog:
- Resaved placeholder normals and properties used in our models. Created 2 normals: blue for hairs and yellow for everything else.

## Tests:

# How to test:
Confirm our attachments like tusks, draenei horns look the same as before.


ValianBlue (approved):
_Did a quick look through all the races and they all seem to have their features appear correctly._

TheGroxEmpire (approved):
Compared both tusks and Draenei horns in master and they look to be identical to this PR.

PR #668, opened 2021-09-09
1st History Corrections

zumbak04:
## Changelog:
- Grom is said to be 46 when he died.
- At the same time, Grom is said to be a few years older than Durotan and Orgrim. Thus their birthdates are changed to be 2 years younger than Grom.

## Tests:

# How to test:
Just check their age is correct on the 603rd date.


IronyOwl (approved):
Yep, Grom born 556, Durotan and Orgrim born 558.

PR #672, opened 2021-09-11
Goblin Body Type Fix

ercarp:
## Changelog:
- Removed `body_average` from goblin ethnicity so they can gain weight.


zumbak04 (approved):
LGTM!

PR #674, opened 2021-09-11
Gene Refactoring

zumbak04:
## Developer changelog:
- Moved mod genes to a separate file.
- `gene_hair_aging` can be applied to DNA.

## Tests:

# How to test:
Check that characters with `gene_hair_aging` look fine.

### DIFF
--- a/common/dna_data/wc_human_dna.txt
+++ b/common/dna_data/wc_human_dna.txt
@@ -3285,6 +3285,7 @@ tirion_fordring_dna = { # by Garthor13
  		static_tusks={ "no_static_tusks" 0 "no_static_tusks" 0 }
  		gene_bs_ear_tip={ "ear_tip_pos" 0 "ear_tip_pos" 0 }
             gene_skin_value={ "skin_value_neg" 0 "skin_value_neg" 0 }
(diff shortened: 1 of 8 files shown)


ercarp (commented):
Works. Tested it by adding `very_fast_hair_aging` to Varian's DNA.

image

</details>

With the previous gene, it was possible to drag the slider to change the rate at which their hair grayed. I realize this wasn't how it actually worked, but if it did work this way, it would make it easier to fine-tune the rate of graying for specific characters.

image

</details>

Just an idea. If it's not possible to get it to work this way, that's alright.

zumbak04:
> With the previous gene, it was possible to drag the slider to change the rate at which their hair grayed. I realize this wasn't how it actually worked, but if it did work this way, it would make it easier to fine-tune the rate of graying for specific characters.

Well, it didn't really work. Like...
At the max-age hairs were more white than they should be.
The slider affected color, not age so I would rather add more genes like this.

ercarp:
> Well, it didn't really work. Like...
> At the max-age hairs were more white than they should be.
> The slider affected color, not age so I would rather add more genes like this.

In that case, I will approve this. Good job. 👍

ercarp (approved):

PR #693, opened 2021-09-15
Female Trigger for Ruler Designer

zumbak04:
## Developer changelog:
- Added `is_female_portrait_trigger` that must be used instead of `is_female` in portrait modifiers since `is_female` doesn't work in the Ruler Designer.

## Tests:

# How to test:
Confirm blood trolls have their markings and have no problems.


ercarp (approved):
Nothing seems to have broken as far as I can tell. Female blood trolls are still using the same markings as before.

TheGroxEmpire (approved):
Can also confirm that Female Blood Troll is working as intended.

PR #740, opened 2021-10-19
Race Refactoring

zumbak04:
# Changelog:
1) Fixed an issue where a dead character could look of the wrong race.
2) Increased the default **_Different Race_** marriage modifier from **_-175_** to **_-200_**.

# Developer changelog:
- Made all the race traits to be a part of the `race` trait group.
- Removed the `race` variable.

## Tests:

# How to test:
To test the 1st bugfix from the changelog, start as a high elf, use `set_culture nerubian` to change your culture, click on yourself, in the **_DEBUG Interactions_** click on **_Have someone Slay them!_**, the character model of the dead character shouldn't change to a nerubian.

Due to refactoring test marrying different races. Marring a nerubian to a gnome or something like this shouldn't be possible.

### DIFF
--- a/common/script_values/wc_race_values.txt
+++ b/common/script_values/wc_race_values.txt
@@ -1,4 +1,4 @@
-﻿neg_attractive_race_marriage_val = -175
+﻿neg_attractive_race_marriage_val = -200
 
 neg_attractive_race_seduce_chance_val = -20
 neg_attractive_race_seduce_will_do_val = -20


KazutoSensei (approved):
I tried to marry as a gnome a nuberian and switch around, it didn't work, so everything is fine in my opinion.

Also I started as a high elf and changed my culture to nuberian, after that I killed myself and I stayed high elf, so well done.

Issue #729, opened 2021-10-05
Rework Size Innovations


zumbak04:
**Describe your suggestion in full detail below:**
The design of size innovation is far from ideal.
<details>
<summary>Click to expand</summary>

![image](https://user-images.githubusercontent.com/46384992/136044014-d0d5dd0e-8523-4240-83c4-ad45ee76ed3e.png)

</details>
These bonuses and nerfs are applied to characters, not provinces.

So, here, as the Lich King, I get really cool commanders having no minuses.

<details>
<summary>Click to expand</summary>

![image](https://user-images.githubusercontent.com/46384992/136044494-5d520218-e0f9-4b18-8431-0513822ecac2.png)

</details>

PR #772, opened 2021-11-22
Rework Sizes

Gaben1025:
## Changelog:
- Rebalanced sizes using maa buff/debuffs #729 

## How To Test:
- Play around and see if any specific culture has an unfair advantage over the other given their stats:

image


common/culture/innovations/wc_cultural_innovations.txt
```diff
@@ -27,9 +27,17 @@ innovation_massive_soldiers = {
 		is_giant_culture_trigger = yes
 	}
 	modifier = {
-		levy_size = -0.5
-		advantage = 20
-		army_maintenance_mult = 0.15
+		men_at_arms_limit = -2
+		men_at_arms_maintenance = 0.12
+		skirmishers_damage_mult = 0.1
+		skirmishers_toughness_mult = 0.08
```
zumbak04, line 33: Don't get why damage and toughness differ.

Gaben1025: Toughness is a more valuable stat, would you prefer they be the same?

Also I will remove the maa limit like we talked about and replace it with more maintenance cost.

zumbak04: > Toughness is a more valuable stat, would you prefer they be the same?

How it's more valuable? Yes.

common/culture/innovations/wc_cultural_innovations.txt
```diff
@@ -63,9 +77,17 @@ innovation_small_soldiers = {
 		is_small_culture_trigger = yes
 	}
 	modifier = {
-		levy_size = 0.3
-		advantage = -10
-		army_maintenance_mult = -0.05
+		men_at_arms_limit = 1
```
zumbak04, line 80: Is it intended?

zumbak04 (changes requested):

zumbak04 (approved):
LGTM!

PR #781, opened 2021-11-30
Silver Hand Order

zumbak04:
## Changelog:
- Made the Knights of the Silver Hand a holy order instead of theocracy in the year 603 and 605.

## Tests:

# How to test:
Confirm it's a holy order.


ValianBlue (approved):
_Changes verified. ✅_

PR #822, opened 2022-01-05
Naked Bugfix

zumbak04:
## Changelog:
- Fixed an issue where goblins spouses converted to `lordaeronian` culture were naked.

## Tests:

# How to test:
Use `set_culture lordaeronian <Character ID>`, the ID is visible when you hover the mouse over the character.


Soltysutine (approved):
The following bug fix seems to do the trick and the goblins no longer become naked and change into the culture's appropriate clothing. I approve of this

PR #824, opened 2022-01-07
Asexual races event

171917:
Set all Asexual Races in the game to be asexual

## Changelog:
- Added event to set asexual races to be asexual

## Developer changelog:
- Created a limit which searches for a specific trait instead of a flag for the trait

## Tests:

# How to test:

### DIFF
--- a/events/wc_events/wc_race_events.txt
+++ b/events/wc_events/wc_race_events.txt
@@ -7,10 +7,27 @@ WCRAC.10 = {
 	
 	trigger = {
 		NOT = { has_character_flag = race_giving_fired_flag }
+		
+		has_race_trait_trigger = no
 	}
 	
 	immediate = {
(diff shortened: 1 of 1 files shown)


events/wc_events/wc_race_events.txt
```diff
@@ -95,4 +95,26 @@ WCRAC.30 = {
 		# Long-lived races slow down their XP
 		set_long_lived_lifestyle_xp_slowdown_event_effect = yes
 	}
+}
+
+
+# Characters of Asexual races are set to Asexual
+WCRAC.40 = {
```
zumbak04, line 102: Instead of creating an event, you could do this check-in `set_race_effect`.

zumbak04 (changes requested):

Issue #841, opened 2022-01-12
Undead Orc Vassal of Kul Tiras


zumbak04:
**Describe your suggestion in full detail below:**
Undead orcs of Crestfall shouldn't be a vassal of Kul Tiras.
<details>
<summary>Click to expand</summary>

![image](https://user-images.githubusercontent.com/46384992/149237967-d104881f-b10c-41a1-871a-bf2d7e2dfcf2.png)

</details>

PR #842, opened 2022-01-13
Removed Orc Vassal of Kul Tiras

Loralius:
## Changelog:
Crestfall is no longer a vassal of Kul Tiras by 603 and 605 bookmarks.

## Developer changelog:
Per #841 

## Tests:


zumbak04 (approved):
LGTM!

Issue #552, opened 2021-07-24
Plunder Isles to Southsea Culture


zumbak04:
**Describe your suggestion in full detail below:**
Southsea culture is very small so it's on the verge of extinction. If we want to prevent it, we have to convert the majority of the Plunder Isles to Southsea. I suggest turning Gurubashi, Hozen lands to Southsea (since they have a strong presence in the world and region anyway.

Remerod:
Could we not add some random Islands to the area between Plunder Isles and Pandaria?. i really like the plunder isles as it is, a bunch on different cultures all grasping for dominance, I personally think it is the most "Sandbox" start in the base mod.

I know we gotten some of the island expedition islands in right? but not all right? even if we have there most be some lore islands or wc3 islands that's not in the mod that we could but there?

<details>
<summary>Click to expand</summary>

Uncharted Islands
![image](https://user-images.githubusercontent.com/79544635/126903657-a1f5c511-3a0f-42b1-9199-5596bd6fa51c.png)

Some Unused but planned islands?
![image](https://user-images.githubusercontent.com/79544635/126903773-94ce617a-ae81-4c7c-b35b-4b66253122d9.png)
![image](https://user-images.githubusercontent.com/79544635/126903787-1888f5f5-70c4-4060-ad76-240c8ac6263f.png)

Wc3 islands not placed on the map might might fit the region
Mystic Isles
![image](https://user-images.githubusercontent.com/79544635/126904157-b85af639-931b-4051-b3db-2390b0ce0da0.png)
Petrified forest
![image](https://user-images.githubusercontent.com/79544635/126904223-5332cd01-1a34-49d8-bb97-4e04f66d805b.png)
Emerald Shores
![image](https://user-images.githubusercontent.com/79544635/126904432-3caccea2-58dd-44a6-ae0d-5a2cf1f9e2f6.png)
Sunrock Cove
![image](https://user-images.githubusercontent.com/79544635/126904470-b262eb62-4473-42cd-bf83-13752b02cb2f.png)
Typhoon
![image](https://user-images.githubusercontent.com/79544635/126904580-51f4b387-7b5a-4341-befc-cb9f590542ee.png)
Hurricane Isle
![image](https://user-images.githubusercontent.com/79544635/126904633-4879532d-407e-410c-ac87-919a26b9f9d0.png)
Mur'gul Oasis (could be part of an island)
[...]

Issue #210, opened 2021-02-26
Northsea Pirate Culture


zumbak04:
**Describe your suggestion in full detail below:**
Northsea pirates should get different culture because they're not southern at all.
It may be either new culture or Solstice culture.

ercarp:
Perhaps they could be based on Asian culture? Blizzard never explained which part of the world they were based on, so it could be a nice way to tie them into the game's lore a bit better. [Chinese pirates](https://en.wikipedia.org/wiki/Category:Chinese_pirates) were a thing.

PR #848, opened 2022-01-15
Northsea Provinces -> Feudal

Loralius:
## Changelog:
- Changed holding type in Sorlof's Strand and Scalawag Point from tribal to feudal. 
- Added tradeport building in Scalawag Point to replace Tribal Market.

## Tests:


history/provinces/00_k_utgarde.txt
```diff
@@ -313,7 +319,7 @@
 706 = {	# Sorlof's Strand
 	culture = pirate_culture
 	religion = tidemother
-	holding = tribal_holding
+	holding = none
```
zumbak04, line 322: None?

history/provinces/00_k_utgarde.txt
```diff
@@ -301,7 +301,13 @@
 705 = {	# Scalawag Point
 	culture = pirate_culture
 	religion = tidemother
-	holding = tribal_holding
+	holding = castle_holding
+	
+	570.1.1 = {
+		buildings = { 
+			common_tradeport_01
```
zumbak04, line 308: Can't say I get why you specified it. Does it come from the lore?

Loralius: From [WoWhead](https://wowpedia.fandom.com/wiki/Northsea_Freebooters) : 
> The Northsea Freebooters are a group of pirates that arrived in Howling Fjord and set up a port at Scalawag Point.

It's only a couple of ramshackle buildings but the port at Scalawag Point is (at least in WotLK) the main base of operations for the Northsea Freebooters, so I'd even argue that this is where the capital of the top title should be. Scalawag Point is where the pirates drop off their booty to be traded or shipped elsewhere.

Sorlof's Strand (where the capital currently is) is much more valuable but it's basically a strech of uninhabited beach in WoW and [Captain Ellis](https://wowpedia.fandom.com/wiki/Captain_Ellis) and his crew are sunk and killed by the Sea Giant Sorlof here so that could be reflected by changing the culture of the county is the capital is shifted?

While we are talking about balance, it is a secondary reason for why I added the tradeport building. As discussed in #552, the Southsea Culture is endangered and I figured this would be a small buff for them until #210 is addressed.

zumbak04 (approved):
LGTM!

TheDrPibb1999 (approved):
All workes for me!

PR #889, opened 2022-02-10
Era dates corrected for every culture

### LINKED ISSUE #583: Era Dates
Tribal and Early Medieval dates are messed up

image

pedromdpereira:
## Changelog:
- Era dates corrected for every culture.

image

Closes #583


Gaben1025 (commented):
Why do all cultures now start their history at the year 100? Im pretty sure the only thing you need was the defines file.

MemeWizard:
> if we plan to have a start date at year 1 every culture will have no history

I don't think we plan it
Neither we can because of technical problems anyway, such as every character being literally 1 years old at game start

MemeWizard (approved):
Looks good to me

PR #900, opened 2022-02-15
GUI optimisation

MemeWizard:
## Developer changelog:
- Replaced all `GuiScope` calls that don't have matching `End` to eliminate memory leak they cause.
- Updated army_disband.dds to fit the vanilla icon size better

## Tests:


zumbak04 (approved):
LGTM!

MemeWizard:
@zumbak04 thanks for review <3

PR #956, opened 2022-09-09
Gnome Goggles

Chomski2:
Basic gnome goggles

## Changelog:
- Added goggles model for gnomes

## To-do list:
- [x] Goggles ingame
- [x] Use the dynamic color palettes
- [x] Female model

## Tests:
- [ ] Only some characters with the gnome trait use the headgear.

# How to test:
Version 1.4.4


Chomski2:
Theres a way to make the lens of the goggles slightly transparent but then the texture of the pattern make a grid effect that I cant find a way to remove or made less visible.

Goggles without alpha
unknown

Goggles with alpha
unknown (1)

ercarp (commented):
I said it before but I'll say it again, these are incredible! They make the gnomes look so much more like gnomes. 😄

My only point of feedback would be to make the lenses transparent, but as you pointed out in a comment above, there seems to be a graphical issue tied to this currently. But if we can find a way around that, then they would be perfect.

If there's no way to do that, I think a good compromise might be to lower the saturation on some of the lenses. They're very faithful to how they appear in WoW, but I feel like the super-saturated colors might clash a bit with the more realistic, toned-down colors of CK3.

I also wonder if it might be better to tie the use of goggles to something else instead of just being random. For example, they could perhaps be used by gnomes with a high Learning stat so the visual would be tied to a gameplay element.

zumbak04 (approved):
LGTM!

PR #1016, opened 2023-03-05
Goblin Banking 

PIKSKING:
## Changelog:
- Added this thing.
- Changed that thing.

## Developer changelog:
- Added this thing.
- Changed that thing.

## Tests:

# How to test:


Gaben1025:
I would just copy the Borrow from holy order decision as best as you could, and add a little change for goblins

Here is the localization:
image
image

Here is the decision:

image

image
image

Gaben1025 (changes requested):
goblo.dds needs to be 1100x440

image

common/decisions/wc_loan_decisions.txt
```diff
@@ -0,0 +1,22 @@
...
+	}
+	is_valid_showing_failures_only = {
+		is_imprisoned = no
+	}
+        
+      sort_order = 0
+
+	effect = {
+		custom_tooltip = get_loan_decision
+		remove_short_term_gold = 5
```
Gaben1025, line 17: I am not sure why they will lose 5 gold.

common/decisions/wc_loan_decisions.txt
```diff
@@ -0,0 +1,22 @@
+﻿get_loan_decision = {
```
Gaben1025, line 1: Make the format, names, effects, descriptions, ai checks, and values all similar to the holy order decision, including the repay part. There only need to be minor tweaks from that decision since all we are essentially doing here is giving an option to those without a holy order. A difference I would also make is that Goblins want to make more money out of it (so either they ask for repayment faster or ask for higher interest, idk if the original decision does that but I'm just giving an example)

localization/english/decisions/wc_get_loan_I_english.yml
```diff
@@ -0,0 +1,6 @@
+l_english:
```
Gaben1025, line 1: As I mentioned, make the descriptions similar to those of the Holy Order description with tweaks to mention Goblins. @Warcraft-GoA-Development-Team/writer / @Warcraft-GoA-Development-Team/english-proofreaders  can help with this if needed.

Gaben1025 (changes requested):

PR #1040, opened 2023-03-16
Goblin Loan Decision

PIKSKING:
This is the Goblins Loans system,Check for bugs or missing properties,also open to any suggestion

## Changelog:
- remade the whole thing.

## Developer changelog:
- remade it

## Tests:

# How to test:
Play as Anyone (not evil relegion or tribal) and see if you can take a loan,then switch to a tribal/evil relegion character and see if the decision is still there.


common/decisions/goblinmoney_decisions.txt
```diff
@@ -0,0 +1,79 @@
...
+	ai_check_interval = 0
+	selection_tooltip = goblin_money_tooltip
+	desc = goblin_money_decision_desc
+	confirm_text = goblin_money_decision_confirm
+
+	is_shown = {
+                      
+                      highest_held_title_tier > tier_barony
+		NOT = {
+                     highest_held_title_tier > tier_barony
```
Gaben1025, line 14: image

you are causing a contradiction, this is why it does not show up at all.

Gaben1025 (changes requested):

common/decisions/goblinmoney_decisions.txt
PIKSKING: Done,test again before release if its possible

TheGroxEmpire (changes requested):
The decision seems to work fine even after switching characters. However, the image proportion on this decision is a bit off. The description also doesn't appear.
<details>

image
image

</details>

PIKSKING:
the image is good for now,till i swtich to another image but the other one is a localization problem

IronyOwl:
Seems a bit rough.

First time the game crashed entirely. Not sure if it was related to this or something else, but it's something to consider if it happens again.

500 gold for 100 prestige feels rather powerful. Granted it's a loan, and 75 in interest is fairly steep, but I wonder if it's strong enough to be a standard strategy.

After a while an event fired giving me a fairly hefty negative trait for not paying back the goblins, but there wasn't enough text to explain just what was happening. Did I take too long to pay it back and they got mad?

Once that was done, I was able to take out another loan, and then use that loan to pay off the first loan, resulting in no loans. Or at least, there was no button to repay the second loan, so I assume it got wiped out along with the first.

There's hardly any localization, flavor text, or an icon for the negative trait.

PIKSKING:
Oh wait,i know the localization problem,its because a wrong tag for the descs
but i couldnt try it in good condition with the mod.

PIKSKING:
I did a remake from the previous works and mods, I am sure this time it should work.

Gaben1025 (changes requested):
image

- 2 events

image

- No image
- You get a bonus from lending even though the description says you are late to pay back the loan
- Goblins do not get payback loan casus beli on loanee

PR #1044, opened 2023-03-19
Update goblin_money_l_english.yml

MoonyMeanie:
Note, gb_1 is meant to represent the Steamwheedle Cartel, and gb_2 is meant to represent the Bilgewater-

## Changelog:
- Added English localization for Goblin Trade Cartel events.


PIKSKING (approved):
I think it looks good,but we can expand the system and make it better in the future.

PR #1118, opened 2023-04-12
Aging penalty fix

sumankun:
## Changelog:
- removed vanilla mortality system, so all became immortal (in `00_defines.txt` file)
- instead of vanilla system we have `character modifiers`, which reduce `health`/`prowess`
- It uses equivalent human age. We apply modifier after >= 30(human) years old. It is updated on every birthday. There are 365 days, and each character has it own birthday, so it doesnt touch performance
- we have two variations of that penalty. One with prowess, one without prowess. It depends on `dynasty legacy` (which removes prowess reduction)
- removed `game start health` bonus, i believe it was used to make them live longer
<details>
<summary>Looks like this</summary>

image
image

</details>

## Data for health penalty calculation:
I wrote vanilla numbers in excel to calculate how fast it reduces health/prowess with vanilla mortality system. Use `10.0 health` and `100 prowess` as starting numbers.
<details>
<summary>Age 25-80</summary>

image
</details>
<details>
<summary>Age 80-120</summary>

image
</details>

- I increased prowess losing chance number a little bit, because vanilla uses preudorandom
- In `modifier file` i added to every modifier comments like `#+0.2`. It means i make health penalty bigger from above excel table by this amount. It is required because game thinks it is illness, so it adds health resistance bonus when detect modifier.

## To-do-list:
- [x] Balance aging penalties, so humanity dies like in vanilla (health resistance bonus is main enemy)
## Tests:
- [x] There are no errors in `changed files` in `Documents\Paradox Interactive\Crusader Kings III\logs\error.log`

# How to test:
1. Start up version `1.8.2` of **Crusader Kings III** and look for human characters using _observer mode_.
2. If human dies as intended, you can check other races too, but they should be ok as long as human race is ok.


ElMariuso (approved):
Everything looks ok, I'm not sure because it's a big change to how Vanilla works, so I think more reviewers are needed.

ercarp (approved):
Observed a game for about 200 years and everything looks good to me. The modifier is correctly applied to all characters when they reach a human equivalent age of 30 and characters seem to be dying of old age between the ages of 60-75 for the most part; I did observe some dying a bit sooner, but not many. The median seems to be around 75 from what I can tell.

However, aside from Duke Newall of Balor (died at 95) and Count Panoramix of Blackwald (died at 88), I don't think I saw a single new human character age past 80, which could be seen as a "problem". I personally feel like humans would be more likely to grow past 80 in the Warcraft universe where you have access to priests, druids, and paladins with healing magic - however, that might be better suited for a different PR.

For now, I would say this work as intended and I can't personally see any problems with the implementation. It's nice that elves aren't dying randomly in their 70s anymore. LGTM!

common/modifiers/wc_age_penalty_modifiers.txt
```diff
@@ -0,0 +1,185 @@
+﻿30_age_penalty = {
```
zumbak04, line 1: Ooof, that's many modifiers. Why you didn't use `stacking = yes` modifier?

common/on_action/wc_aging_penalties.txt
```diff
@@ -0,0 +1,165 @@
...
+}
+on_birthday_aging = {
+	trigger = {
+		30 <= { value = age divide = racial_age_multiplier_value }
+	}
+
+	effect = {
+		if = {
+			limit = {
+				has_trait_with_flag = can_not_get_sick
```
zumbak04, line 14: Check `is_immortal`.

zumbak04, line 14: I would add a new trait flag with its own description. This way you won't need the trait limits below.

common/on_action/wc_aging_penalties.txt
```diff
@@ -0,0 +1,165 @@
...
+			}
+			clear_age_penalties = yes
+		}
+		else = {
+			save_temporary_scope_value_as = {
+				name = my_human_age_multiplier
+				value = racial_age_multiplier_value
+			}
+			
+			if = {
```
zumbak04, line 26: If you implement `stacking = yes` on the modifier, you can replace this else/if logic with a math formula. The input of the formula should be your age, the output should be "stacks" of the modifier.

zumbak04 (changes requested):

common/on_action/wc_aging_penalties.txt
```diff
@@ -0,0 +1,165 @@
...
+			limit = {
+				has_trait_with_flag = can_not_get_sick
+				NOT = { has_trait = being_light }
+				NOT = { has_trait = being_void }
+			}
+			clear_age_penalties = yes
+		}
+		else = {
+			save_temporary_scope_value_as = {
+				name = my_human_age_multiplier
```
zumbak04, line 22: Not sure why it's called `my_human_age_multiplier`. It gets the mutiplayer for all the races, not only humans.

common/on_action/wc_aging_penalties.txt
```diff
@@ -0,0 +1,166 @@
...
+			if = {
+				limit = {
+					aging_trigger_between = { AGE_1 = 30 AGE_2 = 35 }
+				}
+				clear_age_penalties = yes
+				add_age_penalty = { AGE = 30 }
+			}
+			else_if = {
+				limit = {
+					aging_trigger_between = { AGE_1 = 35 AGE_2 = 40 }
```
zumbak04, line 36: Optimization.
If the effect fires, you're 35-40 years old, but do have the penalty modifier. This returns false and you go all the way down else/if chain.

Solution:
You split `aging_trigger_between` in 2 triggers: `has_aging_penalty_trigger = { AGE }` and `race_relative_age_between_trigger = { AGE_1 AGE_2 }`. First you check `race_relative_age_between_trigger`. If true, you check `has_aging_penalty_trigger`. If false, you get the modifier.

common/scripted_effects/wc_age_effects.txt
```diff
@@ -10,4 +10,55 @@
 		}
 		set_age = scaled_age_value
 	}
+}
+clear_age_penalties = {
```
zumbak04, line 14: The effect names should end with `_effect`.

common/scripted_triggers/wc_age_triggers.txt
```diff
@@ -75,4 +75,12 @@ is_child_trigger = {
 is_teen_trigger = {
 	is_child_trigger = no
 	is_adult = no
+}
+aging_trigger_between = {
```
zumbak04, line 79: The trigger names should end with `_trigger`

common/traits/wc_infusion_traits.txt
```diff
@@ -32,6 +32,7 @@
 	health = 5
 	
 	flag = can_not_get_sick
+	flag = no_aging_penalties
```
zumbak04, line 35: Add a localization.

common/modifiers/wc_age_penalty_modifiers.txt
zumbak04, line 1: The modifier names should end with `_modifier`.

zumbak04 (changes requested):

localization/english/triggers/wc_character_triggers_l_english.yml
```diff
@@ -36,6 +36,13 @@
  has_trait_with_flag_can_not_romance_first:0 "You are #bold not#! allowed to romance"
  has_trait_with_flag_can_not_romance_third:0 "[CHARACTER.GetShortUIName|V] is #bold not#! allowed to romance"
 
+ has_trait_with_flag_no_aging_penalties_global:0 "Is #bold not#! aging"
```
zumbak04, line 39: I meant `TRAIT_FLAG_DESC_<flag name>` localization, but this is nice too. Keep it.

zumbak04 (approved):
LGTM! But left one comment to address.

PR #1128, opened 2023-04-16
Orcish clans update

ElMariuso:
## Changelog:
- I renamed the titles in all the files to make it clearer.
- I moved the titles in specific files for them, just to find them more easily (in landed titles, and coat of arms)
- I changed the color of most clans on the map to be less saturated, and more in line with Warcraft 2.
- I added the titles of the missing clans, as well as holder places for the Draenor Horde (which will become the Fel'Horde) and the Dark Horde.
- All clans are prioritized to be the primary titles for the AI. So the Clan Blackrock, does not turn into the Kingdom of Black Morass like that...
- I changed the color of the cultures to make it more appropriate to their clan.
- I changed the name of the Horde to the Orcish Horde. This could change when Thrall reforms the Horde for example.
- Major clans are kingdoms now.
- The Horde flag has been changed to its old symbol.

## Things that could be added to this PR if people want.:
- When you reach a certain level of power, your clan can go from Duchy to Kingdom or from Kingdom to Empire.
- If a clan title is destroyed, someone from the corresponding culture can recreate it.
- A decision to create the Dark Horde.
- That the lands after a war of invasion, are first distributed in a hardcoded way to the clans (to fit with Warcraft 2), before that is for the creation of other vassals.

## Tests:

# How to test:
- Just launch the game and check the changes I made.


ercarp (approved):
I like all of the aesthetic changes (colors, titles, etc.) and don't see any problems with any of them, but I can't speak for any mechanical changes since I'm not too familiar with that aspect of modding. But as far as the pretty colors and titles are concerned, LGTM!

PR #1142, opened 2023-04-23
Ulduar Hill Forts Fix

Gaben1025:
## Changelog:
-Replaced Walls and Towers Level 1 with Level 2 Hill Forts for Ulduar


arithon (approved):
can't check in game atm, but code-wise it looks good to me

PR #1149, opened 2023-04-30
Health penalty fixes

sumankun:
## Changelog:
- added loa blessing and hibernate modifier into health penalty system consideration
- added all races with small life expectancy bonus into health system (like tauren ones)
- added small life expectancy bonuses (from dynasty legacy, some traits and holy sites) into health system
- removed color2 from landed titles, because lazy to create another pr

## Developer changelog:
- blessing loa gives life_expectancy = 200, hibernated gives life_expectancy = 600
so in 270 age and 670 they should be like 70 years old human.
270/70 = multiplier => multiplier = 3,85, we reduce it by 1(defaule human value), we get 2,9
670/70 = multiplier => multiplier = 9,57, we reduce it by 1(defaule human value), we get 8,5
This is what i added

## Tests:

# How to test:


ElMariuso (commented):
It seem to works, but I find that with "Human age equivalent" it's quite ambiguous.
Is there a way to change this to account for hibernations and blessings in this system?

sumankun:
> Is there a way to change this to account for hibernations and blessings in this system?

It is "human equivalent age" including blessing and hibernations. It gives info to player to understand where you are with all your life/age bonuses.

ElMariuso (approved):
Approved

Gaben1025 (approved):

PR #1151, opened 2023-05-06
Two Death CoA + Argent Dawn CoA

Rekutur:
Added two Death CoA Emblems.
scourgeskull&frostmourne1
+ Added Argent Dawn CoA.


TheGroxEmpire (approved):
The CoAs are applied in the game correctly.

PR #1153, opened 2023-05-08
Neutral CoAs added.

Rekutur:
Added 
2 Ogre flags
2 Wolvar flags


TheGroxEmpire (approved):
The CoA is applied in the game correctly.

PR #1175, opened 2023-07-05
Character sexualities

Grekopithikos:
## Changelog:
Many characters who have stated lovers and soulmates now have set sexualities.
Dagran and Moira get married when they become soulmates
The couples Genn and Mia, alongside Nielas and Aegwynn become soulmates when they get married

All Sayaads are now automatically bisexual
Sayaads are more likely to grow up to be lustful

## Tests:

# How to test:
Check if Sayaadi children grow up to be bisexual and most likely lustful

### DIFF
--- a/common/culture/cultures/wc_highborne.txt
+++ b/common/culture/cultures/wc_highborne.txt
@@ -94,7 +94,7 @@ blood_elf = {
 	unit_gfx = { western_unit_gfx }
 	
 	ethnicities = {
-		10 = high_elven
+		1 = high_elven
 	}
 }
 
@@ -125,6 +125,6 @@ nightborne = {
 	unit_gfx = { western_unit_gfx }
 	
 	ethnicities = {
-		10 = high_elven
+		1 = high_elven
 	}
 }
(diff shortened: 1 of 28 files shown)


TheGroxEmpire (commented):
- Aegwynn isn't set to Soulmate with Nielas Aran
<details>

image

</details>

Other than that, it appears to be working correctly.

Grekopithikos:
>Aegwynn isn't set to Soulmate with Nielas Aran

Isn't that because he's dead?

TheGroxEmpire:
> > Aegwynn isn't set to Soulmate with Nielas Aran
> 
> Isn't that because he's dead?

That is true. Curious why would we set up it as Lovers when it doesn't matter because he's dead. But it's a non issue.

TheGroxEmpire (approved):

Grekopithikos:
>That is true. Curious why would we set up it as Lovers when it doesn't matter because he's dead. But it's a non issue.

Well it was mostly for the sake of lore details and just in case we ever get an earlier bookmark

PR #1176, opened 2023-07-07
Age-Fertility hotfix

Rekutur:
Made fertility %-100 after 70.

# How to test:
Start the game, watch the long-living races, see if they still have children after 70 years old.


TheGroxEmpire (commented):
We need to discuss more about this feature. I'd rather have long living races having extremely low chance to reproduce rather than 0% after reaching 70+. It could be problematic due to how easy it is for babies to die in CK3. It is also a lot more lore friendly as long living race could technically have children in lore.

sumankun:
As i remember we already consider human age for fertility trait in other on action.

PR #1178, opened 2023-07-23
Magnataur CoA Update & 4 new CoAs

Rekutur:
Added 10 Magnataur flags.

# How to test:
Start game.
Look at Magnataur CoAs.
Check if the CoAs are randomly generated for Magnataurs.


TheGroxEmpire (changes requested):
Magnataur CoA flags doesn't appear to be generated at all.

TheGroxEmpire (changes requested):
I can now see the CoA in the game. However, some CoA have same color as the background. For example, these CoAs has white color on its emblem on white background:
<details>

image

image

image

</details>

TheGroxEmpire (approved):
My issues are resolved.

PR #1236, opened 2023-10-31
Union rework

sumankun:
## Changelog:
-universal code to easier add new unions. Each union is one story cycle
-You can call your union members in offensive wars only if you are the leader of the union. High (de)centralization can change it
-Marriage penalty modifier for acception doesnt take union allies into account
-Forsaken can easier join horde if they took any kingdom (Lordaeron for Sylvanas). And people dont leave the union if Sylvanas(or other undead) became a leader of the Horde
<details>
<summary>added 4 new unions</summary>

image
</details>

 <details>
<summary>universal decisions for create/join/leave union</summary>

image
</details>

<details>
<summary>new union game rule. If you enable it, all states inside one faction will have same color on political map. Every union has its own color</summary>

image
</details>

 <details>
<summary> union elections. Every candidate has a score. The score is a chance to become the next leader. Your chance = (your score/total score of all candidates). Score depends on various conditions(tier level, stats, prestige level etc). You can invest prestige/gold to increase chances</summary>

image
</details>

<details>
<summary>centralization-decentralization system, which decides what bonuses union members/union leader get from character modifier. It also influences who you can call into offensive war from union and determines penalty/bonus for score of primary heir of the current union leader</summary>

image
</details>

## Tests:


common/game_rules/00_game_rules.txt
Grekopithikos, line 1: Why was this moved?

ElMariuso (approved):
LGTM as I said in PM

PR #1239, opened 2023-11-17
1.11.1 update

Grekopithikos:
## Changelog:
- Update for 1.11.1
- made as many adaptations as possible. Like `invite persian scholars` into `invite dalaranian scholars`. Persian `frontier tradition` made into tradition for blood elves/scarlets. New sassanian sword you could find during travelling in Persia you can find as ancient elven sword either in northern kalimdor or broken isles etc etc etc
- new clan system is fully adaptated, if you make for example horde as clan, it will work, i tested. I saved as many events/decisions for persian clans/tax collectors as possible too. Deleted only those which have heavy persian lore in localization and in triggers.

## Tests:

# How to test:


ElMariuso (approved):
Well, the error.log look good and I don't see any issue in game.
Some errors remain, but do not seem to be related to the update.

LGTM

Issue #1245, opened 2023-11-22
Potential Balancing Issue: the Lich King can be really strong with max dread & the Guile dynasty finisher 


JamarcusRue:
**Your mod version is:**
Steam version release alpha

**What expansions do you have installed?**
Garments of the HRE, Fashion of the Abbasid Court, the Nortehrn Lords, the Royal Court, the Fate of Iberia, Tours & Tournaments, Elegance of the Empire, Legacy of Persia, Wards & Wardens. 

**Are you using any submods/mods? If so, which?**
GoA - Theramore Expansion & Architect of Wonders are the two main game-changing ones. The others are attached.

**Please explain your issue in as much detail as possible:**

**Steps to reproduce the issue:**
1) Load save. You'll be Jaina. Keep the game paused. 
2) Enable debug mode, whether using the mod I was using (F10 then F11) or via other methods.
3) Character switch to the Lich King.
4) Get the Lich King out of bankruptcy. Give him 25 gold. 
5) Select the Archmage. He's in Ambermill. 
6) Offer vassalization. 

**Upload an attachment below: .zip of your save, or screenshots:**

![BugReport6of6](https://github.com/Warcraft-GoA-Development-Team/Warcraft-Guardians-of-Azeroth-2/assets/151694407/3e99fca1-52d2-45bf-bb14-bb661d76a8ff)
![BugReport5of6](https://github.com/Warcraft-GoA-Development-Team/Warcraft-Guardians-of-Azeroth-2/assets/151694407/08a40ae0-8775-4eb3-8595-c57234035975)
![BugReport4of6](https://github.com/Warcraft-GoA-Development-Team/Warcraft-Guardians-of-Azeroth-2/assets/151694407/03da4d6b-1a14-4855-9b43-a8563000a992)
![BugReport3of6](https://github.com/Warcraft-GoA-Development-Team/Warcraft-Guardians-of-Azeroth-2/assets/151694407/412d11fe-68f4-4d9c-b0e5-2a2b958fb7e7)
![BugReport2of6](https://github.com/Warcraft-GoA-Development-Team/Warcraft-Guardians-of-Azeroth-2/assets/151694407/78755b06-ad8d-41df-9840-efa6385c5ba9)
[...]

PR #1248, opened 2023-11-22
DRAFT - Scourge random assign undead (second try)

edmund-h:
## DRAFT PULL REQUEST!
**This is just a suggestion for possible changes, with peace and love ✌️ ➕ ❤️, dont take it seriously and please ping me with any suggestions on how to make it better**

## Changelog:
- Lower acceptance for titles, marriage, vassalization, etc to LK (esp if in Union)
- LK makes people undead by inviting them to court
- LK makes people undead by granting titles
- 90% chance for population in undead culture lands to be undead at game start.

## Developer changelog:
change from previous pull request: 
- added negative modifiers to address issue #1245 
- made the conversion only happen for Lich King because only he has the lich powers as of yet to make people undead

## Tests:

# How to test:

### DIFF
--- a/common/character_interactions/00_courtier_and_guest_interactions.txt
+++ b/common/character_interactions/00_courtier_and_guest_interactions.txt
@@ -534,6 +534,8 @@ invite_to_court_interaction = {
 	desc = {
 		desc = invite_to_court_interaction_desc
 
+		# Warcraft
+		# Warning for Scourge players
 		triggered_desc = { 
 			trigger = { 
 				scope:actor.culture = { has_cultural_pillar = heritage_undead }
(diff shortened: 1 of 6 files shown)


Grekopithikos:
Isn't turning a character undead already an option you can have when giving a title to someone?
Why would the Lich King not be able to control whether they turn undead or not?

edmund-h:
> Isn't turning a character undead already an option you can have when giving a title to someone?
No it only currently happens when you vassalize someone, and it is in fact an option you have
> Why would the Lich King not be able to control whether they turn undead or not?
you’re right, what i will do is make this optional in order to bring it in line with how vassalization works

PR #1253, opened 2023-11-26
Harpy models

sumankun:
## Changelog:
- Removed static harpies
- Added real models
## Tests:


ElMariuso (approved):
I don't see any problem with portraits, no clipping or anything.
And nothing in error.logs either.

The harpies looks good, they look vicious. LGTM

PR #1263, opened 2023-12-05
Playable Governments

sumankun:
## Changelog:
- Theocracy and Republics are playable
- Added Theocratic elections for Kaldorei
- Adde Tyrande into Kalimdor bookmark
- Added CoAs for Tyrande and Kaldorei
- Fixed some stuff from previous PR

## Tests:
- [ ] There are no errors in `wc` files in `Documents\Paradox Interactive\Crusader Kings III\logs\error.log`


ElMariuso (changes requested):
<details>

image
</details>

Change the localization here, if you can. Same for Republics. It's no longer unplayable.

<details>

image
</details>

The majority of goblins should have the republican government too.

<details>

image
</details>

My governement changed at the succession, it shouldn't happen. Can you check that ?

```
[23:57:08][jomini_script_system.cpp:276]: Script system error!
  Error: Event target link 'scope' returned an unset scope
  Script location: file: common/scripted_effects; general_heir_agp_effects line: 66
 file: common/on_action/wc_agp_on_actions.txt line: 82
 file: common/scripted_effects; make_an_heir_hof_agp line: 62
```
I have like that 82 times in my file too, it probably happened at the succession. The scope is probably undefined in succession, if it's used. Moreover, at no time do voters vote for anyone, they all remain at 0 votes. All the candidates the same. If that helps to find the origin of the problem.

sumankun:
> My governement changed at the succession, it shouldn't happen. Can you check that ?

It returns day after death. This is done in janky way

ElMariuso (approved):
LGTM! No more errors in logs

PR #1268, opened 2023-12-07
Draenei / Exodar fixes

edmund-h:
## Developer changelog:
- remove nonworking unbalanced unit counter 
- feudalized draenei holdings on landing.
- reduced the average year for the expdar crash to 614, event should occur no later than 620 and this should be extremely rare. uncommon occurrences should be between 612-618.
- changed localization paths for `exodar.1009` to match the id

## Tests:

# How to test:


edmund-h:
@Gaben1025 requested this change

usfshihy (commented):
Exodar province (b_balaar) should be a city holding probably. Like this: set_holding_type = city_holding

I don't see any other problems, but do keep in mind that your working on an older branch and will need to update it to how the master branch has it.

events/wc_events/wc_exodar_crash_events.txt
```diff
@@ -305,33 +304,100 @@ exodar.1009 = {
...
+								culture = { has_cultural_pillar = heritage_cenaric }
+							}
+						}
+						desc = exodar.1001.c
+					}
+					desc = exodar.1001.d
+				}
+			}
+		}
+		remove_character_flag = viewer
```
edmund-h, line 401: whoever reviews this, please pay special attention to the changes in this file. I'm not sure where they came from as they just kinda popped up on my branch when i tried to make changes to the localization paths for exodar.1009. it looks like they might have been removed in master and i was working off an old branch? i just want confirmation to put them back before i delete anything.

usfshihy: Add this back yes, this removes the flag after the event is over.

usfshihy (approved):
lgtm.

common/scripted_effects/wc_exodar_effects.txt
```diff
@@ -190,34 +190,49 @@ setup_titles_azuremyst_exodar_effect = {
...
+		add_special_building = exodar_01 
+	}
 	title:c_yanra = {
 		set_county_culture = culture:draenei
 		set_county_faith = faith:naaru
 		change_development_level = -2
 		change_county_control = -25
 		add_exodar_crystal_low_modifier = yes
+		title_province = {
+			set_holding_type = castle_holding
```
sumankun, line 204: I think you said you wanted to make those things through `if` statement?

usfshihy: This is true, on the chance the holding is already a castle, then it will throw an error in the error log.

sumankun (changes requested):

common/scripted_effects/wc_exodar_effects.txt
```diff
@@ -190,34 +190,64 @@ setup_titles_azuremyst_exodar_effect = {
 		change_county_control = -50
 		add_exodar_crystal_mid_modifier = yes
 	}
-	title:b_valaar.title_province = { add_special_building = exodar_01 }
+	title:b_valaar.title_province = {
+		if = { 
+			limit = { holding = tribal_holding }
+			set_holding_type = castle_holding
```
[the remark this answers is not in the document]
edmund-h: fixed; that was embarrassing

sumankun (approved):
lgtm!

PR #1270, opened 2023-12-08
Terrain 1.11 fixes

Raged50:
## Changelog:
- Modified following locators so provinces are shown correctly on Kalimdor:

- 5148
- 5050
- 5072
- 5071
- 5044
- 5070
- 2324
- 2325
- 5074
- 5073
- 5060
- 5042
- 5075
- 1626
- 5049
- 5060
- 5024

- Added trees to EK


sumankun (approved):
Both trees and provinces looks good

PR #1286, opened 2023-12-18
Made Naga Government Feudal instead of Eldritch

Myphicbowser:
## Changelog:
- Removed the Void Being trait from all Naga, as suggested by Raged in my suggestion post.

## Tests:
- [X] The mod takes less than 5.5 GB in the Task Manager (Windows)


Myphicbowser:
I would like to have this either be discussed or deleted, I thought the idea was fine enough as the Religion Update will add their new religion and then we can add the Void Being trait to Azshara and have her rule Feudal Governments as an Eldritch. But I do not like PRs just sitting that change simple things and have been discussed.

Raged50 (approved):
LGTM, error log clean

PR #1288, opened 2023-12-19
Dragonflight Government

Myphicbowser:
Creates the Dragonflight Government and gives it to all Dragonflight Titles. Making it technically Opt-Out.

## Changelog:
- Added Dragonflight Government Type.
- Changed all Dragonflight Titles to grant Dragonflight Government Type.

## Tests:

### DIFF
--- a/common/game_concepts/wc_government_game_concepts.txt
+++ b/common/game_concepts/wc_government_game_concepts.txt
@@ -11,3 +11,6 @@ demonic_government = {
 eldritch_government = {
 	parent = dark_government
 }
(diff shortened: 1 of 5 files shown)


edmund-h:
I like this, but i would like to see something unique about it that makes it dragon-y. Preferably, this would be something that helps the dragons defend their far flung territories, that can work for AI as well as human players.

Myphicbowser:
I've elected to not continue with this, as I believe someone else can do far better, and I lack the skills currently to make it happen.

PR #1294, opened 2023-12-24
Arcane Scrying Tenet

edmund-h:
## Changelog:
- Removed the Astrology tenet, as it's underpowered and doesn't represent anything useful in the Warcraft universe.
- Replaced it with the Arcane Scrying trait which provides additional protection from random events and improves relations with other arcane magic users without rising to the level of a syncretism trait
- Removed it from the Lightforged as it is no longer compatible with their magic laws and replaced with Monasticism.
- Replaced all other instances of Astrology with Arcane Scrying, ensuring that magic laws allowed arcane magic.

## Tests:

# How to test:


common/religion/doctrines/00_core_tenets.txt
```diff
@@ -1824,7 +1825,7 @@
 		
 		# Warcraft
 		is_shown = {
-			tenet_moon_worship_allowed_trigger = no
+			always = no
```
Grekopithikos, line 1828: I don't think astrology should be removed completely.
It could be kept as a cheaper alternative or for faiths where arcane scrying doesn't fit(like those who shun the arcane).

common/religion/religions/water_group.txt
```diff
@@ -515,7 +515,7 @@
 			#Special
 			doctrine = special_doctrine_eastern_kingdoms_syncretism
 
-			doctrine = tenet_astrology
+			doctrine = tenet_arcane_scrying
```
Grekopithikos, line 518: This really does not fit tidemother at all.
The reason as to why astrology was added in the first place was because of its naval bonus.
It should have its own tenet, like "Sacred Sailing", greatly boosting naval speed and other water-related modifiers.

common/religion/doctrines/00_core_tenets.txt
edmund-h: would you be OK with renaming it something like "augury" so that it kind of denotes superstitious fortune telling, because astrology itself doesnt really work in the warcraft setting

common/religion/religions/water_group.txt
edmund-h: ok i'm open to this, although it might even be better to make it a unique cultural innovation or something so that jaina can keep it even being a kirin tor worshipper

edmund-h:
this will be moved to Magic 2024

PR #1296, opened 2023-12-29
Give important lore modifier in any bookmark

sumankun:
## Changelog:
- We were not giving important lore modifier correctly, so it worked only in first bookmark. Because we were giving it on birth. Storm event already has trigger to not happen for important lore ones. Now i fixed it, and it give lore modifier for 10 years.

## Tests:

# How to test:
For example open jaina in any bookmark, she always gets it for 10 years


Grekopithikos (approved):
Plor armor now lasts 10 years as intended
LGTM

PR #1305, opened 2024-01-06
fix modifier localization and rename files

edmund-h:
## Developer changelog:
- Fixed localization for a modifier from vanilla that didn't have any for some reason
- Renamed all files from the `forsaken-events` and `forsaken-story-cycle` PRs to `forsaken` rather than `lordaeron`

## Tests:

# How to test:


helloruhe (approved):
thanks

PR #1323, opened 2024-01-07
Racial stats

Grekopithikos:
## Changelog:
- Added stats for the race traits.

## Developer changelog:
- The Eredar trait modifier seem to reflect the draenei traits. The problem with that is that the Man'ari also use the trait so it might be best to just make a new trait for them.

## Tests:

# How to test:


Myphicbowser (commented):
Code works, traits work, we fixed what wonkiness there was (The Murlocs/Pygmies/Makura Health Debuff)

No game breaking issues, makes the game feel more alive, this is my seal of approval as a player and tester

Myphicbowser (approved):

PR #1327, opened 2024-01-12
Mrligion Patron Gods

Myphicbowser:
## Changelog:
- Added Patron Gods for the Mrligion
- Removed Communal Identity in favor of the Patron Tenet

## Tests:

# How to test:
To test it, just make sure the tenet shows up, the modifiers are applied, the event shows up, and there are no missing localizations

# Current Issues
At first posting, the Tenet doesn't show up and the error log fills up REAL quick, it actually isn't that huge but its far bigger than it should be, looks like the biggest issue is that I probably skipped a step that defines shit, but I'll get around to it this weekend


common/religion/doctrines/wc_tenets.txt
```diff
@@ -0,0 +1,45 @@
+	tenet_patron_loa = {
+		name = {
+			first_valid = {
+				triggered_desc = {
+					desc = tenet_patron_loa_name
+				}
+				desc = tenet_patron_loa_desc
+			}
```
Grekopithikos, line 8: I think just using the vanilla bhakti trait would be easier than adding a patron god.
Just add Mrligion in `is_shown` in the triggers and it'll work just fine.

Myphicbowser:
Looks like it was mentioned that this entire religion will be getting re-visited at some point, so for now I'm closing it

PR #1329, opened 2024-01-13
Create wcw_worgen_traits

Afromidwestern:
these are the Worgen traits I created, I will send the localization files and the changes to the plague scheme in a bit.

## Changelog:
- added to traits for worgen, feral worgen and balanced worgen
- Changed that thing.

## Developer changelog:
- Added this thing.
- Changed that thing.

## Tests:

# How to test:

### DIFF
--- a/common/traits/wcw_worgen_traits
+++ b/common/traits/wcw_worgen_traits
@@ -1,0 +1,61 @@
+###################
+# worgen traits
+# by Afromidwestern "mystical dying rat"
+###################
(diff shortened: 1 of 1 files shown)


Raged50:
No need to merge this as worgen aren't currently in development and how the concept/implementation of they look will probably alter in the future once magic is implemented.

I might reach out at a future stage when we're ready to start looking at worgen for your input.

PR #1362, opened 2024-01-28
Removers Plot Armor Modifier

Myphicbowser:
## Changelog:
- Added a new Dynasty Modifier (Thanks Skesus)
- Changed all lore important Modifiers to use the Dynasty Modifier, as its important and passes on to their children, ensuring all the Wyrnns that don't develop a fear of boats live long enough to become important

## Developer changelog:
- Skesus added a new Dynasty Modifier, we manually added it to every Lore Important character's dynasty.
- The Dynasty Modifier is exactly the same as the temporary plot armor, but hooked onto the Dynasty and lasts forever.

## Tests:

# How to test:
Make sure all lore important characters have a Dynasty Modifier that makes them not die to stupid shit like Storms at sea


Grekopithikos:
Wasn't plot armor originally made so that the more significant characters would survive for the next 10 years?
Wouldn't expanding that to a permanent dynasty modifier make things unfair for everyone else?

Raged50:
Agree with Grek, it was to give some canon characters additional protection at the start of the game and that was pretty much it. The idea was that we shouldn't be too reliant on lore characters to achieve things in the game.

Myphicbowser:
Plot armor as a modifier currently makes them safe for fifteen years, that's pretty much it. So is making this a dynasty modifier unfair to others? Yes. However counter point, starting as any of the lore characters already is inherently unfair. People want to see these lore characters and their dynasties for a long time on the game, and since almost all of them have a family size of two and many are long lived races that reproduce slowly, the modifier ends and then we quickly run out of Stormrages and Blackhands, no one sees them anymore and that's what's dissapointing.

PR #1366, opened 2024-02-11
Cleaned up error log

Chomski2:
## Changelog:

Cleaned up errors from the log

## Tests:
- [ ] There are no news errors on `Documents\Paradox Interactive\Crusader Kings III\logs\error.log`


Raged50:
LGTM no errors

PR #1387, opened 2024-04-24
Fix all nations spawning without legitimacy

robmart:
## Changelog:
- Fix all nations spawning without legitimacy

## Developer changelog:
- The piece of code giving all nations legitimacy on game start was not moved over to the mod, this fixes that
- Republic and theocratic nations don't have legitimacy as a mechanic

## Tests:

# How to test:
* Check that everyone has legitimacy


helloruhe (approved):
lgtm

PR #1396, opened 2024-06-03
dark gov can build hospices

Raged50:
# How to test:

Select either Necro or Demon character and construct hospices


edmund-h:
is this the fix we really want? it doesnt seem thematic for dark governments to build hospices

edmund-h:
different fix added in #1397

PR #1400, opened 2024-06-23
Evil Religious Attitude

ElMariuso:
## Changelog:
- Changing the "By Any Means" special doctrine, who give bonus to evil religions, into a pluralism doctrine (To replace fundamentalist for example) for technical purposes.

## Tests:


localization/english/religion/wc_core_tenets_l_english.yml
```diff
@@ -64,10 +64,10 @@
  consumerist_opinion_active_county_opinion_name:0 "Hated Consumerist County"
  doctrine_parameter_opinion_of_consumerist_opinion_active:0 "[opinion|E] of Consumerists: $VALUE|+=0$"
  INCOMPATIBLE_TENET_CONSUMERISM_TRIGGER:1 "Incompatible with the [core_tenet|E] [GetFaithDoctrine('tenet_consumerism').GetName( GetPlayer.GetFaith )]"
-
- special_doctrine_end_justifies_means_name:1 "By Any Means Necessary"
- special_doctrine_end_justifies_means_desc:0 "While others think about what is right and what is wrong, we act. Our goal is righteous. Our will is unshakable. From the ashes shall arise a new order that will shake the very foundations of the world."
  
+ doctrine_pluralism_dark_name:0 "Dark"
```
edmund-h: Can we make this a little more interesting of a title, like "Forces of Darkness"?

ElMariuso: It can be.
I'm just worried about the localization.
Like “We are a Forces of Darkness religion!”

edmund-h: you don't have to worry about that, because the localization for the doctrine is not used that way. "Eastern Conclave" for isntance is something we made up. no one in azeroth would ever say that

PR #1412, opened 2024-08-07
Scaling Knowledge Overflow

robmart:
## Changelog:
- Knowledge Overflow is now a scaling modifier, instead of a stacking one
  * According to test, this should make the game run at least 5% faster

## Developer changelog:
- Should reduce lag upon gaining many stacks of Knowledge Overflow

## Tests:

# How to test:
- Get loads of perks `add_<lifestyle>_lifestyle_xp 100000` 
- Wait a day
- Open character panel
- Check that the Knowledge Overflow modifier has been applied correctly, and that there is little/no lag
- Do the same thing in the `dev` branch and compare how laggy it is


helloruhe (approved):
amazing

PR #1440, opened 2024-09-08
Fix union block error spamming

robmart:
## Developer changelog:
- Fix union block spamming the error log because wrong effect to remove characters from global variable list was used

## Tests:

# How to test:


ElMariuso (approved):
good good

PR #1458, opened 2024-09-16
Arcane portal ritual 

helloruhe:
## Changelog:
- Add Arcane Portal Ritual
   - Rank 1: Teleport a target to specific location
   - Rank 2: Teleport multiple characters
   - Rank 3: Teleport an army + commander
- Todo:
  - [x] Add ability to cast even when other rituals are on cooldown
  - [x] Teleport travel entourages
  - [x] Can't teleport army in combat tooltip
  - [x] Army battle warning tooltip
  - [x] Add decision / event to come back to title location
 
## Developer changelog:
- This ritual does not trigger the cooldown of other rituals
- The general ritual cooldown has been shortened

## Tests:

# How to test:


FelModder:
✅: Working as Intended
❌: Not Working as intended
🟡: Practically working but has bugs/issues Working

**Arcane Portal Ritual**
❌ Rank 1: Teleport a target to specific location
(i casted it on my neighbouring character and then used switch character to switch to the victim to see if it works)
Reasons: The effect of the spell told me that it would take atleast 6 months before i could use the decision but i could use it immediately. (_Also, is this really a debuff? the character doesn't really go to the destination because it doesn't avail me a travel decision to go back to my land. That's where the true debuff of the spell lies in my opinion. Basically, no option of teleport travel entourage that you have listed above in "Todo" list._) 
Arcane Portal Rank 1

      
❌ Rank 2: Teleport multiple characters (wtf is that UI lmao, Screenshot no. 1)
Reasons: Just not letting me cast it despite none of the selected characters being in combat/battle. Yes, the destination has been properly selected. (Screenshot no. 2)
Arcane Portal Ritual Rank 2
Arcane Portal Ritual Rank 2 not working


✅ Rank 3: Teleport an army + commander
(Working as intended) (The AI just gives up trying to bring their army back home and just lets the teleported army stay where they are, may need fixing but not necessary right now)

PR #1504, opened 2024-10-11
TB Error fix

Malicos:
## Developer changelog:
- Modified TB base modifier to not cause errors.

### DIFF
--- a/common/modifiers/wc_event_modifiers.txt
+++ b/common/modifiers/wc_event_modifiers.txt
@@ -328,7 +328,7 @@ wc_thorium_base_modifier = {
 	icon = county_modifier_development_positive
 	levy_size = 0.25
 	levy_reinforcement_rate = 0.25
-	army_maintenance_mult = -0.2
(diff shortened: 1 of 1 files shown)


common/modifiers/wc_event_modifiers.txt
```diff
@@ -328,7 +328,7 @@ wc_thorium_base_modifier = {
 	icon = county_modifier_development_positive
 	levy_size = 0.25
 	levy_reinforcement_rate = 0.25
-	army_maintenance_mult = -0.2
-	men_at_arms_recruitment_cost = -0.25
+	travel_danger = -10
```
robmart: This could maybe be swapped to `fort_level`? I think that would be more thematic

robmart (changes requested):

common/modifiers/wc_event_modifiers.txt
Malicos: swapped

robmart (approved):

PR #1573, opened 2024-11-07
Resurrected Characters Skills Fix

robmart:
## Changelog:
- Resurrected characters will now have the same skills as they did prior to their death, instead of randomly assigned ones

## Tests:

# How to test:


FelModder:
✅ "Resurrected characters will now have the same skills as they did prior to their death, instead of randomly assigned ones" is working as intended. I have tested this with three different characters and found no changes in their **base** skills after resurrecting them.

FelModder (approved):

helloruhe (approved):

PR #1618, opened 2024-12-28
Plague of Undeath

ElMariuso:
## Changelog:
- The scourge plague now uses the in game plague system 
   - Some of the black death events are now used for the scourge 
   - Added a few unique events for the scourge plague

## Tests:

# How to test:
- Open the game


FelModder:
✅: Working as Intended
❌: Not working as Intended
🟡: Is working but has minor issues

(Since i know there an actual Plague scheme "Ritual Spell Spread Plague" available for the Death Magic Spell Tree, i'm concerned if this makes the Lich King too powerful... Rather i'm questioning if this plague of undeath should only be a special Magic spell that only the Character holding the "Lich King" trait can use i.e Lich King and Arthas only should be allowed to use)

- Characters are getting converted to the undead trait after they die because of the "Plague of Undeath" Trait ✅ (maybe make it also so that they change their religion to Mordutari (Deathbringer) after they become undead? Would work well into the narrative of the plague being spreaded and causing massive chaos and undead armies getting bigger🟡)

- The Plague is limited to only the Empire Title of Lordaeron ✅ (if more such plagues can be done they need to be limited to Kingdom/Empire titles like this one has been done)

- Dark Frenzy has stopped working in this branch... I killed several randomly prisoned characters and it didn't increase my frenzy at all ❌

- Several Localization are "WIP": ❌
1st Pic: Nothing there for event when the plague spreads to your Title
2nd Pic: Nothing there when your character dies and becomes undead
3rd Pic: This seems to be another pop up when the plague is finally subsiding? Can't put my finger on why and when it does appear (this event is a good one for my first suggestion can be implemented in this one as a choice like "Join the Lich King" or "Believe in the holy light and fight" or A suicide option)
Event Pop up needs localization
No localization for after turning into undead
What

events/scheme_events/wc_plague_scheme/wc_plague_outcome_events.txt
```diff
@@ -1,371 +1,202 @@
...
-		name = wc_arthas_story_9_1_OPT_B
-		trigger = {
-			can_be_in_coalition_trigger = yes
-			NOT = { is_in_coalition_against_trigger = { CRISIS_HOLDER = title:e_scourge.holder } }
-		}
-		join_crisis_coalition_effect = { CRISIS_HOLDER = title:e_scourge.holder }
-		ai_chance = {
-			base = 0
-			join_coalition_scripted_modifier = { CRISIS_HOLDER = title:e_scourge.holder  }
-		}
```
Grekopithikos: Keep in mind this part was deleted when updating to dev

helloruhe: thanks ill add it back

PR #1621, opened 2025-01-03
Misc/Channeled Spells

helloruhe:
## Added 3 New spells (all channeled):
- ### Death Magic
- **Realm of Necrosis:** Adds a modifier to the target that increases dark frenzy gain and decreases decay; if they have a government that uses dark frenzy
- ### Light Magic
- **Power Word: Barrier:** Adds a modifier that increases MR and health resistance
- ### Shadow Magic
- **Dispersion:** Adds a modifier that makes -50% less chance of dying to non-natural causes, but has small health penalty. (Also costs a decent amount of mana)

- The channeled spell type will be used for upcoming branches (life ;fel)
- Channeled spells cost mana per month rather than an upfront sum. Some will cost more mana than you regen.
- Channeled spells don't count towards the spell casting cap (for now)

## Tests:

# How to test:


FelModder:
✅: Working as Intended
❌: Not Working as intended
🟡: Working as described but has MAJOR issues

(Screenshots posted below are only for ❌ or 🟡)
(You will need to majorly focus on how the calculations for the channeling of spells are being made. Most bugs are from gaining too much mana or losing too much when i cast a channeling spell or cancel a channeling spell.)

### Death Magic -> 🟡
- **Realm of Necrosis:** Adds a modifier to the target that increases dark frenzy gain and decreases decay; if they have a government that uses dark frenzy.
(_Screenshot 1: The spell is working as intended._)
(_Screenshot 2: When i go to cancel the spell, it starts taking away more mana from me instead of going back to normal. I was **losing 3.31 mana per month while i was channeling this spell** but **started losing 8.31 mana per month after cancelling the spell**._)
Realm of Necrosis 1
Realm of Necrosis 2 A Bug when Cancelling


### Light Magic -> 🟡
- **Power Word Barrier:** Adds a modifier that increases MR and health resistance.
(_Magic Resistance is increased & health resistance is increased and is properly lost when i stop channeling the spell as intended._)
(_Rank 2 Bug: When i casted the Rank 2 on my wife and myself, **only my wife got the modifier and i didn't get any modifier**. As shown in the first screenshot below._)
(_Rank 2 Bug: When i cancelled the spellcasting, **my wife didn't lose the modifier as she is supposed to after cancellation of the spell**. As shown in the second screenshot below._)
Light Magic Channeled
Power Word Barrier Rank 2


### Shadow Magic -> 🟡
- **Dispersion:** Adds a modifier that makes -50% less chance of dying to non-natural causes, but has small health penalty. (Also costs a decent amount of mana)
(The Dispersion modifier is gained and lost after channeling and cancellation of the spell, respectively. So the modifier is working well and is properly lost when i stop channeling the spell.)
(_Screenshot 1: When i casted the spell, **i started gaining 26.19 mana per month instead of losing it**._)
(_Screenshot 2: When i cancelled the channeled spell, **i started gaining a further amount of 65 more mana per month**._)
Dispersion 1 Bug when casting
Dispersion 2

PR #1638, opened 2025-01-31
fix/l10n/flowers

Fladle:
Changes:
Rose -> Mageroyal
Orchid -> Kingsblood
Aster -> Peacebloom

Notes/question:
In  `get_pressed_flower_species_effect` the region specific flowers for example poppy have the line `flower_poppy_region_trigger = no`. I assumed this meant they wouldnt appear. But if that's not the case i can change those too.

## Tests:


FelModder:
Unrelated to our dev team + only flags/keys have been changed, not the actual Localizations

PR #1682, opened 2025-03-23
Revert "Plague of Undeath (#1618)"

edmund-h:
This reverts commit eef0efb7d0795d0cbaa20c4ac5683ec28dd0435c.

## Changelog:
- Reverts plague update so 1.15 can go ahead.
\

## Tests:

# How to test:


FelModder:
Why is bro still here in 583...
(the plague of undeath is gone and other normal plagues are functioning well, don't think anything needs changing)
why is he still here

robmart (approved):

edmund-h:
> Why is bro still here in 583... (the plague of undeath is gone and other normal plagues are functioning well, don't think anything needs changing) 

this was a pre existing bug that people thought would be too difficult to fix. as it is the way the mod was before the merge, we can reasonably say that fixing it is out of scope for this pr. it will be fixed in my fixes to the plague redo on its own branch

PR #1688, opened 2025-03-30
Fix wanderer events

edmund-h:
## Changelog:
- Added more localization to geo regions
- All new targeting mechanics and localization for "Dreams of Distant Lands" event
- Added relevant support files that thing.

## Tests:

# How to test:


Grekopithikos (approved):
approved

PR #1697, opened 2025-04-01
Flavourized Perks

FelModder:
## Changelog:
- Added different Rewards for "True Ruler Perk" (Diplomacy Tree) as an Evil Government
- Added different Rewards for "Writing History Perk" (Diplomacy Tree) as an Evil Government
- Added different Rewards for "Dignitas Perk" (Diplomacy Tree) as an Evil Government
- Added Warcraft MAA types to some martial perks : Pincer Tactics, Organized March, Hit and Run
- Added "World Weary" version of the "Know Yourself" perk for Undead and Elder Races allowing them to commit suicide in a flavorful way when a player feels they have run their course (AI will never do this).

## Developer changelog:
- This branch is meant to change some perks to be more falvourized for Different governments or Long Living Races (Example: Evil Factions/Governments don't use legitimacy. Hence, the perks that provide legitimacy are useless.)

## Tests:

# How to test:
Open the game.


edmund-h (changes requested):
we should discuss this in dev chat

common/lifestyle_perks/00_diplomacy_2_majesty_tree_perks.txt
```diff
@@ -327,6 +342,7 @@ dignitas_perk = {
 	
 	government_character_modifier = {
 		flag = government_is_landless_adventurer
+		flag = evil_government
```
edmund-h: dark govt doesnt use legitimacy so this is useless for them

common/lifestyle_perks/00_diplomacy_2_majesty_tree_perks.txt
```diff
@@ -337,6 +353,7 @@ dignitas_perk = {
...
```
edmund-h: this is a lot for evil govts and isnt really thematic, they are more about dread. but making it `_per_dread` might make these characters too powerful. we should talk about balancing this

common/lifestyle_perks/00_learning_1_medicine_tree_perks.txt
```diff
@@ -145,6 +145,12 @@ know_thyself_perk = {
 			text = know_thyself_perk_effect
 		}
 	}
+	# Warcraft
+	culture_character_modifier = {
+		parameter = elder_culture
+		parameter = is_undead
+		stress_loss_mult = 0.20
```
edmund-h: on the one hand, i see what you are trying to do here. on the other, stress is one of the few game mechanics that could really hurt an elder culture as a mitigating factor, so i'm not sure this is exactly the kind of solution we need.

common/lifestyle_perks/00_learning_1_medicine_tree_perks.txt
```diff
@@ -208,6 +214,20 @@ healthy_perk = {
...
+		parameter = is_undead
+		health = -0.5
+		epidemic_resistance = -5
+		negate_prowess_penalty_add = 10
+		negate_diplomacy_penalty_add = 4
+		negate_martial_penalty_add = 4
+		negate_stewardship_penalty_add = 4
+		negate_intrigue_penalty_add = 4
+		negate_learning_penalty_add = 4
+	}
```
edmund-h: why does taking this perk nerf your stats? what am i not getting here?

common/lifestyle_perks/00_martial_1_strategy_tree_perks.txt
```diff
@@ -176,6 +184,9 @@ sappers_perk = {
 		pikemen_siege_value_add = 0.1
 		archers_siege_value_add = 0.1
 		skirmishers_siege_value_add = 0.1
+		# Warcraft
+		massive_infantry_siege_value_add = 0.2
+		massive_aerial_siege_value_add = 0.3
```
edmund-h: definitely DO NOT buff massive MAAs. theyre kinda OP already

otherwise these changes are kinda okay, i would go without one or the other buffs to light aerial

common/traits/00_traits.txt
```diff
@@ -1679,6 +1679,18 @@ whole_of_body = {
...
+	culture_modifier = {
+		parameter = elder_culture
+		parameter = is_undead
+		health = -0.5
+		fertility = -0.2 # Remove/Change this later when the fertility system for Long Living Races get's Reworked
+		stress_gain_mult = 0.2
+		prowess = 12
+		martial_per_piety_level = 1
+		learning_per_piety_level = 1
+	}
```
edmund-h: health and fertility dont really matter for undead. do we really need to punish elder cultures more? they already cant have kids most of their life

common/lifestyle_perks/00_diplomacy_2_majesty_tree_perks.txt
FelModder: Do you know how the code works for lifestyle?... Invert_check = yes means it will give legitimacy to those EXCEPT the landless and evil governments. Atleast open the game and see how it works because it does.

common/lifestyle_perks/00_diplomacy_2_majesty_tree_perks.txt
FelModder: Nope. Wrong.
Normal governments get 0.30 prestige per legitimacy level. Which is equal to 1.5 prestige at max legitimacy.
I have now made it so that an Evil government will get 0.015 per dread. Which is equal to 1.5 prestige at max Dread.

common/lifestyle_perks/00_learning_1_medicine_tree_perks.txt
FelModder: The "Carefree Perk" in the Learning Medicine Tree reduces the stress gain by 20% and i have made "Know Thyself Perk" in the Learning Medicine Tree give you stress loss by 20%. So it balances out when i removed it from the Whole of Body Lifestyle Perk.

common/lifestyle_perks/00_learning_1_medicine_tree_perks.txt
FelModder: You are not getting the fact that, above this modifier, the normal perk will give you health = 0.5 and epidemic_resistance = 5. All i did in this modifier was negate those advantages and instead get other advantages.

The code doesn't automatically replace the stuff once you specify what the elder_cultures should get. You need to tell it to remove the old advantages too.

In short, if i remove "health = -0.5" and "epidemic_resistance = -5" from the code above. It will give the elder_culture the old advantage AND the new advantages.

common/lifestyle_perks/00_martial_1_strategy_tree_perks.txt
FelModder: This is only for the siege progress.
Which is canonically true for warcraft 3, that massive aerials were good at sieging.

+ Plus Raged has reviewed these balances in the Testing channel.

common/traits/00_traits.txt
FelModder: Learning Perk for Elder Races 3

common/lifestyle_perks/00_learning_1_medicine_tree_perks.txt
edmund-h: could have phrased this in a less patronizing way but that makes sense and the whole of body one too. btw don’t resolve conversations when someone requests changes.

common/lifestyle_perks/00_learning_1_medicine_tree_perks.txt
edmund-h: can you just get rid of stress loss and add some other benefit please. for example instead you could make it negate part of the penalty from suicide or something. that’s needed for LL characters according to our players.

common/lifestyle_perks/00_diplomacy_2_majesty_tree_perks.txt
edmund-h: it’s your job to defend your code. if you make a calculation you have to explain that to someone who asks. and it would be better if you could do it in a way that isn’t insulting or disrespectful.

common/lifestyle_perks/00_diplomacy_2_majesty_tree_perks.txt
edmund-h: i’ve never even seen `invert_check` before this and if someone misses something you can point it out without talking down. booting up the game is not standard procedure for code review.

common/lifestyle_perks/00_learning_1_medicine_tree_perks.txt
FelModder: This is the kindest reply i have given you in this whole thread lol. I am just trying to explain what you had missed.

common/lifestyle_perks/00_learning_1_medicine_tree_perks.txt
FelModder: Penalty from suicide? can you elaborate?

Elves and other LL races are very rare breeders and stay with their partners for thousands of ages. Negating the penalties if you partner commits suicide or dies in combat is wayy too overpowered than just a stress loss: +20%. Plus, stress loss: +20% is only useful if you hold activities/events related to stress loss like a feast or hiking.

So no? This isn't a bad buff for LL races and they can still get easily stressed out.

common/lifestyle_perks/00_diplomacy_2_majesty_tree_perks.txt
FelModder: When was i disrespectful/insulting here? I have very clearly explained the calculation and that's all i have done in this case, especially.

I matching the energy uk? You say throw a lot of "No." or "No. Don't do this." very oftenly. Yet, you don't see me complaining about you way of communication and tone you take with me.

common/lifestyle_perks/00_diplomacy_2_majesty_tree_perks.txt
FelModder: You should, preferably, comment after you have not only checked the code, but also checked in-game because you clearly missed out most stuff and how i did it.

I always check the code, then play test, and only then comment. I have furthered my discipline when it comes to that. Again, in this whole case you have clearly misread the code for most of the time so an in-game test would have helped and would help me more to check if there are crashes or what not.

common/lifestyle_perks/00_learning_1_medicine_tree_perks.txt
edmund-h: it is a bad buff for them. we all agree that stress should be a mitigating factor for long lived races. 

i’m not talking about negating penalties if your partner commits suicide. i’m talking about eliminating the randomness and the penalties to piety and dynasty prestige that come from making a suicide attempt. if you feel it’s fair then have it only work above a certain age, for instance 100 or so. it’s something players have asked for. nobody asked for more stress loss

common/lifestyle_perks/00_diplomacy_2_majesty_tree_perks.txt
edmund-h: i did check the code. that’s what this is, a code review. just because i missed one thing doesn’t mean you get to insult me.

edmund-h (approved):

PR #1695, opened 2025-04-01
New religions

Grekopithikos:
## Changelog:
- Added a lot of new faiths such as:
- The Sabahi faith in the new Noor religion. It's an Islam-like faith followed by the Wastewander
- Rush'atali, the dark troll faith
- Ebon Blade
- Xavism (icon made by @kaldrskegg)
- Black Harvest, for warlocks not aligned with the Burning Legion
- Azsharism
- A Pirate faith
- Added a gorloc Oracle religion as was suggested in #1359 and #1218
- Faiths with the Chi Pacifism tenet now see each other as astray
- Added lots of custom faith icons
- Moved the Elune's marks parameter from the Moon Worship tenet to the Gender Stratification doctrine
- Added Theramore as a holy site for Holy Light, Kirin Tor and Tirani

## Developer changelog:
- Added this thing.
- Changed that thing.

## Tests:
- [ ] There are no errors in `wc` files in `Documents\Paradox Interactive\Crusader Kings III\logs\error.log` except `portrait_decals.cpp:101`
- [ ] The mod takes less than 5.5 GB in the Task Manager (Windows)

# How to test:

PR #1699, opened 2025-04-03
Legion Necromancy Religion [DRAFT]

edmund-h:
## Changelog:
- **Adds a new religion to the Doombringer group for necromancers and undead aligned with Sargeras.** This is an overlooked transitional religion that would mostly be on the map between Second War and Frozen Throne, except Deadwind Pass which seemed to have stayed out of the Lich King's influence.
- **Changed some characters, mostly those in Deadwind Pass but a few others as well, to the new religion**
- **In the above case, this replaced Maldraxxi**, which is a religion oriented towards the Shadowlands and thus is kind of a generic "death worship" that belies the connections of most associated characters currently in the mod to Sargeras. Overall I do not think Maldraxxi is relevant to any of the bookmarks we currently have, but I think the most pressing problems are resolved here (see justification)
- Made a purple version of the sargerite logo to serve as the religion’s icon 

## Justification:
- **Sets up continuity within the upcoming First War story/lore content feature** and will create a seamless transition from Sargeras-influenced Medivh to Dark-Government Moroes. 
- **Will prevent necromancers on the map from getting steamrolled** by the Horde or other big players; allowing them to make alliances/join unions/get vassalized to help them last until the Scourge comes along. It also fills a potential gameplay option gap for players who want to RP as a Lich or other necromancer on the side of the Burning Legion once we implement that content. 

## Needed Feedback:
In order of priority:
- **How should we localize the religion?** I have some ideas but definitely open to suggestions.
- **Would a death syncretism tenet work for this faith?** it was my original idea to have this tenet but i second-guessed it because it might allow for lore-breaking alliances (ex: Morroes and Lich King). however looking at @Grekopithikos [religion PR](https://github.com/Warcraft-GoA-Development-Team/Warcraft-Guardians-of-Azeroth-2/pull/1695) where he made Xavism fel-syncretic i am reconsidering that it might be acceptable
- I would like to add sanctioned false conversions to all "public evil" religions (ck2 trigger). right now it's a tenet that's just taking up space in nearly all of them and preventing them from getting needed flavor.  what do we think of this? (not necessarily on this branch, might be out of scope, could be better in @Grekopithikos broad-scope[ religion branch](https://github.com/Warcraft-GoA-Development-Team/Warcraft-Guardians-of-Azeroth-2/pull/1695))
- Eventually I would like to give this religion a mechanic to create limited undead plagues, like Forsaken will get (not in scope for this PR, but does this get you hype for this religion?).

## Tests:

# How to test:


Grekopithikos:
I noticed that you changed the religions of characters like Morbent Fel, Morgan Ladimore and Abercrombie, even though they were never stated to be working for the Legion. I understand that this way they will have more chance to survive but it doesn't make much sense for a lore perspective.

From a lore point of view, it makes the most sense for them to remain Maldraxxi, at least until a new necromantic faith for them is made.
Although Maldraxxi in game is used as a blanket faith for all the necromancers/undead who aren't affiliated with the Scourge, Legion or Forsaken, rather than a faith for those from the Shadowlands faction(absolutely nobody in the mod).

PR #1708, opened 2025-04-21
Arthas Pop-up Fix

FelModder:
## Changelog:
- Arthas Expedition against the scourge will actually show up as an activity pop up, notifying the player that they can do it start arthas' story cycle

## Developer changelog:
- Added this thing.
- Changed that thing.

## Tests:

# How to test:


sumankun (approved):
lgtm

PR #1799, opened 2025-12-11
Compatibility for 1.18.2

FelModder:
## Changelog:
- List of files and folders updated can be found here: https://steamdb.info/patchnotes/21120590/

## Developer changelog:
- If any files are drastically altered, note them down here.

## Tests:

# How to test:


common/scripted_effects/00_death_management_effects.txt
```diff
@@ -32,7 +32,7 @@ silent_disappearance_ai_effect = {
 				root = { is_ai = yes }
 			}
 
-			death = {
+			override_death_effect = {
```
Grekopithikos, line 35: Should the disappearance deaths be overrideable?

FelModder: Many other disappearance deaths already had those, so I thought I should just do it for them all. Definitely doesn't make sense but i did this because other such cases in the code already had them overridden.

PR #1810, opened 2026-01-19
Feat/code/leper

Icelandbal:
## Changelog:
- Added decision for Trogg invasion of Gnomeregan and events for the war and different outcomes
- Added things for Leper Gnomes (culture, religion, trait, character interaction)

## Developer changelog:
- Decision to invade Gnoomeregan for Troggs
- New man-at-arms for Trogs
- After Gnomes lose war they can release gas
- New Culture and Religion for Leper Gnomes
- New tradition for the Leper culture
- Option for leper Gnomes to forcefully irradiate characters in prison
- 3 decisions for neutralising modifiers on provinces affected by gas release for Gnomes

## Tests:


common/religion/religions/glowbound.txt
```diff
@@ -0,0 +1,289 @@
...
+		{ name = "holy_order_the_glowing_crusade" }
+		{ name = "holy_order_army_of_poison" }
+		{ name = "holy_order_servants_of_atom" }
+	}
+
+	holy_order_maa = { teutonic_knights }
+
+	localization = {
+		#HighGod
+		HighGodName = GLOWBOUND_NAME
```
Grekopithikos: The localization ought to change to be more like vanilla CK3. I know other religions use the same format but those are holdovers from CK2.
Look at the Shath'gral and Tyrist religions for reference.

common/religion/religions/glowbound.txt
```diff
@@ -0,0 +1,289 @@
+﻿glowbound = {
```
Grekopithikos: I don't think the religion and faith both having the exact same name is a good idea, because then they'd share the exact same description, which might be confusing, as in CK3 a religion is a collection of similar faiths.
Something like radiationism_religion or toxicism_religion for the religion maybe?

common/scripted_triggers/wc_being_triggers.txt
```diff
@@ -6,6 +6,7 @@
 		has_trait = being_life
 		has_trait = being_order
 		has_trait = being_light
+		has_trait = leper
```
Grekopithikos: You should change this to the irradiated trait you made

history/titles/00_k_gnomeregan.txt
```diff
@@ -47,6 +51,10 @@ c_tinkerboom = {
 	18.1.1={
 		holder=dr_gearshaft
 	}
+	604.9.15={
+		holder=16001 #Sicco[10001]
+		liege=0
+	}
```
Grekopithikos: Remove this, it'll cause errors in history

history/titles/00_k_gnomeregan.txt
```diff
@@ -62,6 +70,10 @@ c_wrenchport = {
...
```
Grekopithikos: same here

gfx/portraits/trait_portrait_modifiers/00_trait_modifiers.txt
```diff
@@ -2019,3 +2018,63 @@ magic_good = {
...
+				morph = {
+					mode = modify
+					gene = gene_eyebrows_fullness
+					template = no_eyebrows
+					value = 0.5
+				}
+			}
+		}
+	}
+}
```
Grekopithikos: Maybe move this to a new file since it's not vanilla?

TheGroxEmpire (changes requested):
- Since the Troggs need to besiege multiple counties to get the war score needed to win the war, it is very likely for thermalplugg to be in prison during the war. It is not breaking the events per say but it'd break the event continuity. 

<details>

Image

</details>

- This event is poorly explained. When I clicked it I felt as if the event did nothing to me. It should explain what exactly it will do.

<details>

Image

</details>

- There are two events that do the same / similar thing and the only difference is that the one cost prestige and the other one cost gold. It should be merged into single event and the player can choose if they want to use prestige or gold.

<details>

Image

</details>

- If Gnomeregan county was for some reason got claimed by other than the Leper or Gnomes, it'd skip the cleaning phase of the event and I can fire Operation Gnomeregan immediately.

<details>

Image

Image

</details>

localization/english/event_localization/wc_troggs_invasion_events_l_english.yml
```diff
@@ -0,0 +1,33 @@
...
+ trogg_invasion.0004.option:0 "We should never have trusted him"
+ trogg_invasion.0004.optionb:0 "Maybe his way of ruling is better..."
+ trogg_invasion.0005.t:1 "Thermaplugg's offer"
+ trogg_invasion.0005.desc:0 "The forces of the troggs may be too much for us to handle. Fortunately, Mekgineer Thermaplugg has proposed a plan that might be our only chance of surviving. According to him, we should leak radiation into Gnomeregan, and the contamination should be able to deal with the troggs. It may have severe consequences for us, but it is better than total defeat."
+ trogg_invasion.0005.option:0 "It might help us"
+ trogg_invasion.0005.optionb:0 "We can't trust him"
+ trogg_invasion.0006.t:1 "Strange gas in Gnomeregan"
+ trogg_invasion.0006.desc:0 "We were so close to defeating those damned gnomes, but as we began capturing their capital, strange gas started to spread through the city of Gnomeregan. It quickly wiped out most of our army. Fortunately, it seems the gnomes did not expect it to be so deadly, and many of their own defenders perished as well."
```
TheGroxEmpire:
```suggestion
 trogg_invasion.0006.desc:0 "We were so close to defeating those damned gnomes, and as we began capturing their capital, a strange gas started to spread through the city of Gnomeregan. Rather than wiping us out, most of our warriors shambled through the fumes, growing enraged, with some even seeming empowered by its effects. However, this came at a cost, as many became uncontrollable, turning into a frenzied force that could not be directed. It appears the gnomes did not expect the consequences, as many of their own defenders perished in the chaos as well."
```

This should be more accurate according to the lore on the [wiki](https://warcraft.wiki.gg/wiki/Invasion_of_Gnomeregan):
"Rather than securing the death of the troggs, most shambled through the gas, becoming enraged, and some were even empowered by it."

gfx/interface/illustrations/decisions/decontamination_gnomeregan.dds
Grekopithikos, line 1: I noticed that these illustrations have more MBs than the other ones.
Did you make sure to save them properly?
The CK3 Graphical DDS FAQ says that they should be like this: `Decisions: .dds, 1100x440, Format: DXT1, no minimaps`

TheGroxEmpire (approved):
The issues I mentioned has been fixed and the events worked as intended.

PR #1829, opened 2026-03-13
Index update

Chomski2:
## Changelog:
Created a new gene for custom headgear
Created a new gene for custom clothes
Created a new gene for custom legwear
Created a new gene for custom cloaks

Fixed error with highborne clothes
Fixed glowing iris not working as intended

## Tests:
- [ ] There are no new errors 


edmund-h (changes requested):
Image 

dwarves are still not spawning with beards. 


Screenshot 2026-03-29 at 11 41 33 PM

additionally, please ensure that this update is not causing these error log errors

edmund-h (approved):

PR #1833, opened 2026-03-14
Legion Commander and Fel Orc

FelModder:
## Changelog:
- Added a Fel Orc trait instead of Fel Orcs being Orcs+Demon
- Fel Orcs will use the Demonic Government
- Added an "Embrace Fel Energies" decision for Orcs with Sargerism, Corruptism and Black Harvest faiths. This allows Orcs to become Fel Orcs. You need a demon liege or a demon prisoner for this to work
- Fel Orcs, if they don't have Berserker trait, they will get it via an event after a few months pass
- Added a Legion Commander trait
- Like the conqueror trait, Legion Commander can only be gained at random after passing some checks that are done every 5 years in the game.
- Those eligible for Legion Commander trait: Demons, Fel Orcs and Fallen Demon Hunters. You NEED to have a doombringer religion, like Sargerism and Corruptism. You are independent and a landed ruler
- A Legion Commander will lose the trait if: They become incapable, no longer independent, no longer landed, if you stay peaceful for 10 years, prestige level falls to disgraceful, 3 years or more into debt, lost more than 5 wars

## Developer changelog:
- This branch will introduce two new traits necessary to continue Burning Legion Content

## Tests:

# How to test:


Grekopithikos:
Should fel orc be a race trait? Wouldn't it make more sense if it was a Fel Corrupted trait?

PR #1840, opened 2026-04-20
Orcish Clan Government

FelModder:
## Changelog:
- Evil Governments will no longer have powerful vassals, since all they really do is be annoying about not having a council position
- Evil Government's AI will now be more warlike
- Evil Governments can now raze holdings for gold

## Developer changelog:
- Orcs are not suited for tribal government and neither clan government.
- Orcish Government will be only for Orc cultural heritage and will be sort of a combination between Tribal Government and Nomadic Government
Tribal Government
Nomadic Government

## Tests:

# How to test:


common/flavorization/wc_title_holders.txt
```diff
@@ -258,7 +258,7 @@ warlord_male_orc_king = {
 	flavourization_rules = {
 		top_liege = no
 	}
-	governments = { feudal_government clan_government tribal_government }
```
Grekopithikos, line 261: Not sure about removing these because in the rare instance where orcs are somehow feudal/clan/tribal they'll be called the generic vanilla titles.

FelModder: Yeah, makes sense. I fixed this.

PR #1867, opened 2026-07-29
Arthas Expedition Refactoring

ElMariuso:
## Changelog:
- This branch is essentially a large suggestion. I'm not certain that everything I've done is correct. That's why an huge code or performances review would be appreciated.
- If things were better before, no worries. I will just delete this PR.
- While developing *Path of the Damned*, I found that managing the consequences and the aftermath of the Arthas expedition was quite complicated. Looking at the code, I realized that many things could be improved and optimized to make handling it easier.
- I also noticed that the activity was generating a huge number of errors in the error.log, so I tried to clean that up as best I could.
- I think the plague content I made before leaving should be back.

## Developer changelog:
- I removed the majority of the global variables that the activity had created.
- Localization now uses scopes instead of constantly looking for global variables, making it a bit more readable and dynamic.
- I tried to remove useless part of the code, like parts that were never used (Like certain scopes save)
- I fixed errors from error.log

## Tests:

# How to test:
- Launch the game as Arthas
- Play every path of the Arthas expedition
- Make sure nothing is broken in the features
- Make sure everything is optimized
- There should be almost no errors left now


helloruhe (dismissed):
Good changes

edmund-h (approved):
my suggestions to change global vars back the way they were have been addressed so i approve

PR #1916, opened 2026-09-13
Magic system performance

Ryalois:
Various performance improvements to spells, ai spellcasting and locked regions.

First some more gameplay impacful changes that I find necessary for performance.
 - AI characters regenerate mana every 3 months for triple the value. This is probably the most controversial change here, but with how expensive updating the mana variables is it is basically necessary. Honestly a quarterly or even bi-yearly to yearly mana regen tick could even become the new standard for players as well in my humble opinion.
 - Spells and spell cooldowns progress in 5 day increments. The bars loading less smoothly is more than worth it.
 - Removed extended family and courtiers without coucil or court positions from recommended target list for AI.
 - Locked regions are reworked to be tracked as character variables and evaluated quarterly and on domicile move. Up to two month of inaccuracy feels perfectly acceptable to me personally. The check could even happen every month and it would still be faster than previous implementation.

The rest of the changes are purely optimization of preexistent logic.

### DIFF
--- a/common/on_action/game_start.txt
+++ b/common/on_action/game_start.txt
@@ -16,6 +16,9 @@ on_game_start = {
 		set_global_variable = { name = is_game_start_date days = 1 }
 		# Warcraft
 		check_available_government_game_start_effect = yes
+		every_living_character = {
+			update_locked_region_effect = yes
+		}
 		# End of Warcraft
 		
 		#Epidemics 10 years grace
--- a/common/scripted_triggers/wc_magic_spells_triggers.txt
+++ b/common/scripted_triggers/wc_magic_spells_triggers.txt
@@ -2356,6 +2356,109 @@ spell_is_helpful_trigger = { # To the recipient(s)
 	}
 }
 
+spell_in_class_is_helpful_trigger = { # To the recipient(s)
+
+    trigger_if = {
+        limit = { $CLASS$ = flag:order }
+        OR = {
+		    $SPELL$ = flag:conjure_refreshment
+            $SPELL$ = flag:counterspell
+            $SPELL$ = flag:scry
+		    $SPELL$ = flag:haste
+		    $SPELL$ = flag:arcane_dome
+            $SPELL$ = flag:evocation
+            $SPELL$ = flag:fire_shield
+            $SPELL$ = flag:frost_armor
+            $SPELL$ = flag:frostfire_bolt
+        }
+    }
+    trigger_else_if = {
+        limit = { $CLASS$ = flag:disorder }
+        OR = {
+            $SPELL$ = flag:summon_demon
+            $SPELL$ = flag:spell_stones
+            $SPELL$ = flag:singe_magic
+            $SPELL$ = flag:shadow_bulwark
+            $SPELL$ = flag:dark_presence
+        }
+    }
+    trigger_else_if = {
+        limit = { $CLASS$ = flag:life }
+        OR = {
+            $SPELL$ = flag:natures_cure
+            $SPELL$ = flag:hotw
+            $SPELL$ = flag:wildgrowth
+            $SPELL$ = flag:regrowth
+            $SPELL$ = flag:invigorate
+            $SPELL$ = flag:dreamwalk
+            $SPELL$ = flag:druid_forms
+            $SPELL$ = flag:ironbark
+            $SPELL$ = flag:motw
+            $SPELL$ = flag:invoke
+            $SPELL$ = flag:force_of_nature
+        }
+    }
+    trigger_else_if = {
+        limit = { $CLASS$ = flag:death }
+        OR = {
+            $SPELL$ = flag:army_of_the_dead
+            $SPELL$ = flag:coagulation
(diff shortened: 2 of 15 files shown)


common/scripted_triggers/wc_magic_spells_triggers.txt
```diff
@@ -2356,6 +2356,109 @@ spell_is_helpful_trigger = { # To the recipient(s)
...
 
+spell_in_class_is_helpful_trigger = { # To the recipient(s)
+
+    trigger_if = {
+        limit = { $CLASS$ = flag:order }
+        OR = {
+		    $SPELL$ = flag:conjure_refreshment
+            $SPELL$ = flag:counterspell
+            $SPELL$ = flag:scry
+		    $SPELL$ = flag:haste
```
helloruhe: Haste is special, helpful when recipient is self and harmful when not

common/on_action/game_start.txt
```diff
@@ -16,6 +16,9 @@ on_game_start = {
 		set_global_variable = { name = is_game_start_date days = 1 }
 		# Warcraft
 		check_available_government_game_start_effect = yes
+		every_living_character = {
+			update_locked_region_effect = yes
+		}
```
helloruhe, line 21: If we are going to make this only update every few months, then we need to add a tooltip somewhere when unlocking an isolated region (and on the events) that it may or may not actually become un isolated until a few months later. Because otherwise I will get bug reports on how they got a desolation event but they can't interact with said region right away. 

You also need to check if the de isolation events actually work (particularly Suramar interacting with other places in their event) with this approach

helloruhe: Also, if someone is teleported to an isolated region, and then tagged as isolated, they won't be able to interact with anyone even if they teleport back right away for a few months, so we need to address that as well

common/scripted_triggers/wc_region_triggers.txt
```diff
@@ -574,18 +574,75 @@ flower_snapdragon_region_trigger = {
...
+			#}
+			#trigger_else = {
+				#is_in_$REGION$_trigger = no
+			#}
+		}
+	}
+	trigger_else = {
+		always = no
+	}
+}
```
helloruhe, line 644: We need to make sure we are still using the custom_description or people are not going to understand why they can't interact with said character.

helloruhe (changes requested):
Honestly I appreciate this work, I have not tested it myself/reviewed the error log but this should help performance a lot. Just want to make sure all the edge cases are covered

common/scripted_triggers/wc_region_triggers.txt
Ryalois: This custom_description actually still works. When there are no triggers inside it essentially defaults to always = yes.

helloruhe: Ty

common/on_action/game_start.txt
Ryalois: I tested it out and you can only teleport within a locked region. You cannot teleport in or out of one. The only way to move in and out is moving domicile as a landless adventurer (which is already covered), or moving your capital if you somehow hold land both within and wihout which shouldn't happen anyway. I guess there could still be an edge case with guests or prisoners from raids.

helloruhe: Got it, just need a description or tooltip still though

common/scripted_triggers/wc_magic_spells_triggers.txt
Ryalois: Haste as a helpful spell was already there in the release version I only regrouped what was already there. Will fix though

common/on_action/game_start.txt
Ryalois: I just made it run the region check immediately in region opening events.

common/scripted_triggers/wc_magic_spells_triggers.txt
```diff
@@ -3639,7 +3637,12 @@ satisfy_spell_in_class_triggers_character = {
     in_diplomatic_range_trigger = { target = $TARGET$ }
 
     trigger_if = {
-        limit = { $CLASS$ = flag:order }
+        limit = { 
+            OR = {
+                $CLASS$ = flag:order
+                $CLASS$ = flag:Order # Why are there two order flags with different capitalization?
```
helloruhe, line 3643: because the base game localized the word order and it broke everything loll 😭

helloruhe (commented):
thak you

common/scripted_triggers/wc_magic_spells_triggers.txt
Ryalois: I spent 30min checking why eligible only doesn't work for arcane spells before I noticed the capitalization difference.

helloruhe: omg

helloruhe:
Sorry for the delay. When I get a chance I'll run the game with these changes to make sure there are no runtime errors. I'll just fix anything that comes up. Tysm for this ❤️

